PR #38, opened 2022-07-28
Fix bug when reading of staxml from a directory

savardge:
Hi, 

hope it's okay to suggest code changes with a pull request. There is a bug when using S0B and choosing to read StationXML response from a directory. In stats2inv(), since glob.glob() returns a list, L292 and 293 in noise_module fail. In preprocess_raw() I added a check on if the response is empty (which previously happened when stats2inv failed to read the response from staxml). Also moved the check on empty data in S0B (lines 208-210) earlier because it tr is empty, there will be an indexing error raised at Line 204.

Thanks for making NoisePy available to the community! It's very useful for my work.

Genevieve


mdenolle (approved):
this fixes a bug found when there are multiple stationXML in the file.

PR #44, opened 2022-11-21
Whitening update

lermert:
Hi developers,

when I recently looked in detail at the whitening routine from NoisePy, I noticed that the tapered section is tapered separately from the whitened section of the input spectrum.

This has no consequence for the case where phase-only is chosen, because the amplitude of the main section is 1 and the amplitude of the taper where it connects to the main (untapered) part of the spectrum is also 1. However, for the case where spectral smoothing is used, this may introduce a jump in the fourier amplitude spectrum that may lead to high-frequency artefacts in the time-domain output.

I created an updated routine. This may have to be edited (for example, I used numpy rfftn rather than scipy). I also added a test to check that at least in the part where the signals are expected to be consistent, their respective error is small. Please, after review, remove the plotting statements in this test so that the test can be run without interaction.

### DIFF
--- a/src/noise_module.py
+++ b/src/noise_module.py
@@ -5,7 +5,7 @@ import obspy
 import scipy
 import time
 import pycwt
-import pyasdf
+#import pyasdf
 import datetime
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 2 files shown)


lermert:
P.S.: I suggested tentatively Xiaotao and Chenxin and reviewers, but if they do not have time or someone else has more time, feel free to change that!

mdenolle (approved):
looks great! thanks for adding the test

PR #51, opened 2023-03-07
Remove hard-coded home directories

carlosgjs:
Using `os.path.expanduser('~')` to resolve the user's home directory instead of hard-coding.


mdenolle (approved):
looks clean!

PR #54, opened 2023-03-09
CLI entry point and download step testing during CI

carlosgjs:
The goal of this PR is to enable CI testing of the download step (other steps to follow). The main change is the addition of a command line interface (CLI) and a small refactoring of ```S0A_download_ASDF_MPI.py``` to allow the CLI to call it. With this change, it's easier to execute the steps by passing arguments on the command line instead of changing the code. E.g.:
```
usage: noisepy.py download [-h] [--path PATH] --start START --end END
                           [--stations STATIONS] [--channels CHANNELS]
                           [--inc_hours INC_HOURS]

optional arguments:
  --path PATH           Directory for downloading files
  --start START         Start date in the format: %Y_%m_%d_%H_%M_%S
  --end END             End date in the format: %Y_%m_%d_%H_%M_%S
  --stations STATIONS   Comma separated list of stations or '*' (default)
  --channels CHANNELS   Comma separated list of channels
  --inc_hours INC_HOURS
                        Time increment size (hrs)
```

Currently I only made configurable the parameters needed to test the download, but others can be added too.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -29,4 +29,4 @@ jobs:
         pip install "."
     - name: Test Download
       working-directory: ./src
-      run: python noisepy.py download --start 2019_02_01_00_00_00 --end 2019_02_01_02_00_00 --stations ARV --channels BHZ --inc_hours 1 --path .
+      run: python noisepy.py download --start 2019_02_01_00_00_00 --end 2019_02_01_02_00_00 --stations ARV --inc_hours 1 --path .
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -27,3 +27,6 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install "."
+    - name: Test Download
+      working-directory: ./src
+      run: python noisepy.py download --start 2019_02_01_00_00_00 --end 2019_02_01_02_00_00 --stations ARV --channels BHZ --inc_hours 1 --rootpath .
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -29,4 +29,4 @@ jobs:
         pip install "."
     - name: Test Download
       working-directory: ./src
-      run: python noisepy.py download --start 2019_02_01_00_00_00 --end 2019_02_01_02_00_00 --stations ARV --channels BHZ --inc_hours 1 --rootpath .
+      run: python noisepy.py download --start 2019_02_01_00_00_00 --end 2019_02_01_02_00_00 --stations ARV --channels BHZ --inc_hours 1 --path .
--- a/src/S0A_download_ASDF_MPI.py
+++ b/src/S0A_download_ASDF_MPI.py
@@ -61,11 +61,11 @@ freqmax   = 2                                                   # note this cann
 
 # targeted region/station information: only needed when down_list is False
 lamin,lamax,lomin,lomax = 32.9,35.9,-120.7,-118.5               # regional box: min lat, min lon, max lat, max lon (-114.0)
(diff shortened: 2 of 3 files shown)


src/S0A_download_ASDF_MPI.py
```diff
@@ -77,229 +74,244 @@
 ##################################################
 # we expect no parameters need to be changed below
 
-# time tags
```
carlosgjs, line 80: No functional changes in this block, it all simply got put into the ```download()``` function below and hence every line was indented.

.github/workflows/test.yaml
```diff
@@ -26,4 +26,7 @@ jobs:
     - name: Install project
       run: |
         python -m pip install --upgrade pip
-        pip install "."
\ No newline at end of file
+        pip install "."
+    - name: Test Download
+      working-directory: ./src
+      run: python noisepy.py download --start 2019_02_01_00_00_00 --end 2019_02_01_02_00_00 --stations ARV --channels BHZ --inc_hours 1 --path .
```
carlosgjs, line 32: This downloads <300k worth of data. Is this sufficient for testing? Too much? too little?

src/S0A_download_ASDF_MPI.py
```diff
@@ -77,229 +74,244 @@
...
+
+    tt1=time.time()
+    print('downloading step takes %6.2f s with %6.2f for preprocess' %(tt1-tt0, tp))
+
+    comm.barrier()
+    if rank == 0:
+        sys.exit()
+
+# Keeping this for backward compatibility, but the preferred entry point is: noisepy.py download
+if __name__ == "__main__":
```
carlosgjs, line 316: Do we need to keep this? If we can remove it and update the README/tutorials to use the noisepy CLI we can simplify the code further.

.github/workflows/test.yaml
niyiyu: Maybe having a three-component downloaded? `--channels BHE,BHN,BHZ`. But the wildcard "?" should work for [obspy.client](https://docs.obspy.org/packages/autogen/obspy.clients.fdsn.client.Client.get_waveforms.html). I am not sure why the code says
```
default_chan_list = ["BHE","BHN","BHZ"]                                             
# channel if down_list=false (format like "HN?" not work here)
```

src/S0A_download_ASDF_MPI.py
niyiyu: I am good with that.

carlosgjs: Sounds good, I'll remove it once we have all the steps in the CLI so that I can update the README

niyiyu (commented):
I think you might still need to specify at least "BH" since there are also other channels like EHZ, SLN. Actually, a station may have tens of channels.

.github/workflows/test.yaml
carlosgjs: Done!

carlosgjs:
> I think you might still need to specify at least "BH" since there are also other channels like EHZ, SLN. Actually, a station may have tens of channels.

I removed the argument to the CLI, which will use the default. I.e. ```default="BHE,BHN,BHZ"```. Does that work?

niyiyu:
yes, sounds like `default="BHE,BHN,BHZ"` would be better.

carlosgjs:
> yes, sounds like `default="BHE,BHN,BHZ"` would be better. After this commit, this PR should be considered ready for merge? @carlosgjs

Yes, it's ready now. The default was already in place (see `noisepy.py:35`), so by removing the `--channels` argument in `test.yaml` that's what it will use.

niyiyu (approved):

PR #55, opened 2023-03-10
Edits to S1 array reshaping 

nrich20:
Removed hard coded directories. In S1, I had to do some array reshaping that involved concatenating an empty array to an array with data. I did some preliminary testing, and it seems that this method doesn't affect the data but not entirely sure. The lines to look at in S1 are 306-308, 333-334, and 369-371.


carlosgjs:
@nrich20 , can you update the title of the PR to something a bit more descriptive? Since it'll become part of the repo's history.

src/noise_module.py
```diff
@@ -1331,31 +1331,30 @@ def taper(data):
...
-
-@jit(nopython = True)
-
+#jit(nopython = True)
 def moving_ave(A, N): ## change the moving average calculation to take as input N the full window length to smooth
     '''
     Alternative function for moving average for an array.
     PARAMETERS:
     ---------------------
-    A: 1-D array of data to be smoothed
```
carlosgjs, line 1342: @niyiyu , are these changes needed given that you added `moving_ave_2D` in [this commit](https://github.com/mdenolle/NoisePy/commit/854cefa37625eabef7d3cb12166332b7e85e1ff4)?

src/S1_fft_cc_MPI.py
```diff
@@ -46,7 +49,9 @@
 ########################################
 
 # absolute path parameters
-rootpath  = '/Users/chengxin/Documents/SCAL'                                # root path for this data processing
+
+#rootpath  = '/Users/Nikus/OneDrive/Documents/Work/NoisePy'                  # root path for this data processing
```
carlosgjs, line 53:
```suggestion
```

src/S1_fft_cc_MPI.py
```diff
@@ -413,6 +427,9 @@
 print('it takes %6.2fs to process step 1 in total' % (tt1-tt0))
 comm.barrier()
 
+sfile = rootpath+ '/CCF/2016_07_01_00_00_00T2016_07_02_00_00_00.h5'
```
carlosgjs, line 430: Should this  be `cc_h5` from lines 375:389?
```
cc_h5 = os.path.join(CCFDIR,tname)
```
In general this hard-coded path will fail

src/S1_fft_cc_MPI.py
```diff
@@ -369,7 +382,8 @@
 
             #---------------keep daily cross-correlation into a hdf5 file--------------
             if input_fmt == 'h5':
-                tname = tdir[ick].split('/')[-1]
+                tname = tdir[ick].split('\\')[-1]
+
```
carlosgjs, line 386: To avoid platform specific paths:
```suggestion
                tname = os.path.split(tdir[ick])[1]

```
But please test this first :)

src/S1_fft_cc_MPI.py
```diff
@@ -193,7 +198,10 @@
 
     # get the tempory file recording cc process
     if input_fmt == 'h5':
-        tmpfile = os.path.join(CCFDIR,tdir[ick].split('/')[-1].split('.')[0]+'.tmp')
+        filename = os.path.basename(tdir[0])
+        filename_noext = os.path.splitext(filename)[0]
+        tmpfile = os.path.join(CCFDIR, filename_noext + '.tmp')
+        #tmpfile = os.path.join(CCFDIR,tdir[ick].split('/')[-1].split('.')[0]+'.tmp')
```
carlosgjs, line 204:
```suggestion
```

src/S0A_download_ASDF_MPI.py
```diff
@@ -1,305 +1,313 @@
...
+
+            #if flag:
+            print(ds,new_tags);print('downloading data %6.2f s; pre-process %6.2f s' % ((t1-t0),(t2-t1)))
+
+tt1=time.time()
+print('downloading step takes %6.2f s with %6.2f for preprocess' %(tt1-tt0, tp))
+
+
+sfile = rootpath + '/RAW_DATA/2016_07_01_00_00_00T2016_07_02_00_00_00.h5' 
+plotting_modules.plot_waveform(sfile,'CI','BLC',0.01,0.4)
```
carlosgjs, line 309: @mdenolle @niyiyu , do we want this script to generate the plots? If so, @nrich20 can you use the actual file name generated vs hard-coding?
But if this is not required during download we can just remove these two lines.

src/S0A_download_ASDF_MPI.py
```diff
@@ -1,305 +1,313 @@
...
-            #if flag:
-            print(ds,new_tags);print('downloading data %6.2f s; pre-process %6.2f s' % ((t1-t0),(t2-t1)))
-
-tt1=time.time()
-print('downloading step takes %6.2f s with %6.2f for preprocess' %(tt1-tt0, tp))
-
-comm.barrier()
-if rank == 0:
-    sys.exit()
+import sys
```
carlosgjs, line 1: @nrich20 do you know why this entire file is showing up as different? Were there actual changes? If not you could revert to have a cleaner diff. In this state it's very hard to see if anything actually changed.

carlosgjs (changes requested):

src/noise_module.py
niyiyu: Yeah since I've had the 2D moving average added (sorry directly in the main branch. my bad). And original `moving_ave` should only take care of 1D case. We may not change this function.

src/S0A_download_ASDF_MPI.py
niyiyu: I think the file name is having a problem (a missing `n` in `download`). This change has been made in main branch.

src/S0A_download_ASDF_MPI.py
mdenolle: i do not think we need the plotting actually, this would be just for sanity check but i think that users would do this on their own. Maybe we can have one tutorial/notebook that shows how to plot the data from the various data flow steps (raw data, outputs from S1, and outputs from S2). so i would suggest to remove 308-309

src/S1_fft_cc_MPI.py
```diff
@@ -294,7 +295,9 @@
 
             # load fft data in memory for cross-correlations
             data = source_white[:,:Nfft2]
-            fft_array[iii] = data.reshape(data.size)
+            data= data.reshape(1, data.size)
+            filler=np.empty((1,data.size))
+            fft_array[iii] = np.concatenate((data,filler),axis=1)
```
carlosgjs, line 300: @niyiyu @mdenolle 
At this point `fft_array[iii].shape = (72000,)`, but `data.shape = (36000,)`, so the assignment fails. Padding with 36000 `0s` works, ie avoids the crash, but is this the right thing to do? Did this code used to work?
I wonder if could be a mismatch of parameters since these dimensions are derived out of params such as `cc_len, step, samp_freq`, etc.

src/S1_fft_cc_MPI.py
```diff
@@ -319,7 +322,8 @@
         t0=time.time()
         #-----------get the smoothed source spectrum for decon later----------
         sfft1 = noise_module.smooth_source_spect(fc_para,fft1)
-        sfft1 = sfft1.reshape(N,Nfft2)
+        sfft1_halved= sfft1[0:(int(sfft1.size/2))]
+        sfft1 = sfft1_halved.reshape(N,Nfft2)
         t1=time.time()
```
carlosgjs, line 327: Here `sfft1.shape==(72000,0)`, so it can't reshaped into `(N=4,Nfft2=9000)`

src/S1_fft_cc_MPI.py
```diff
@@ -1,17 +1,16 @@
...
-import datetime
 import os, glob
 import numpy as np
 import pandas as pd
 import noise_module
 from mpi4py import MPI
 from scipy.fftpack.helper import next_fast_len
-import matplotlib.pyplot  as plt
+
+
```
carlosgjs, line 13:
```suggestion
```

carlosgjs (approved):

PR #56, opened 2023-03-13
Update tutorial to use requirements.txt and noisepy.py

carlosgjs:
- Remove `S0A_download_ASDF_MPI.py` as an entry point
- Update README to use the new entry point and to use `requirements.txt` for installation
- Moved defaults to the CLI and display them in the help text:
```
usage: noisepy.py download [-h] [--path PATH] --start START --end END
                           [--stations STATIONS] [--channels CHANNELS]
                           [--inc_hours INC_HOURS]

optional arguments:
  --path PATH           Directory for downloading files (default:
                        /Users/carlosg/Documents/SCAL)
  --start START         Start date in the format: %Y_%m_%d_%H_%M_%S (default:
                        2016_07_01_0_0_0)
  --end END             End date in the format: %Y_%m_%d_%H_%M_%S (default:
                        2016_07_02_0_0_0)
  --stations STATIONS   Comma separated list of stations or '*' for all
                        (default: *)
  --channels CHANNELS   Comma separated list of channels (default:
                        BHE,BHN,BHZ)
  --inc_hours INC_HOURS
                        Time increment size (hrs) (default: 24)

```

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,12 +23,6 @@ Jiang, C. and Denolle, M. "NoisePy: a new high-performance python tool for seism
 # Installation
 The nature of NoisePy being composed of python scripts allows flexible package installation, which is essentially to build dependent libraries the scripts and related functions live upon. We recommend using [conda](https://docs.conda.io/en/latest/) or [pip](https://pypi.org/project/pip/) to install the library due to their convenience. Below are command lines we have tested to create a python environment to run NoisePy. Note that the test is performed on `macOS Mojave (10.14.5)`, so it could be slightly different for other OS. 
 
-## Pre-requisites
-
-An MPI installation is required. E.g. for macOS using (brew)[https://brew.sh/] :
-```sh
-brew install open-mpi
-```
 
 ### Note the order of the command lines below matters ###
 
@@ -36,9 +30,16 @@ brew install open-mpi
 ```bash
 conda create -n noisepy python=3.8 pip
 conda activate noisepy
+conda install -c conda-forge openmpi
 pip install -r requirements.txt
 ```
+
 # With virtual environment:
+An MPI installation is required. E.g. for macOS using [brew](https://brew.sh/) :
+```sh
+brew install open-mpi
+```
+
 ```bash
 python -m venv noisepy
 source noisepy/bin/activate
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Jiang, C. and Denolle, M. "NoisePy: a new high-performance python tool for seism
 * adding a jupter notebook for generating response spectrum for a nodal array (to be done)
 
 # Installation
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -21,33 +21,31 @@ Jiang, C. and Denolle, M. "NoisePy: a new high-performance python tool for seism
...
+```bash
 python -m venv noisepy
 source noisepy/bin/activate
-pip install wheel h5py numpy numba pandas pycwt jupyter mpi4py pyasdf
-git clone https://github.com/mdenolle/NoisePy.git
+pip install -r requirements.txt
 ```
 To run the code on a single core, open the terminal and activate the noisepy environment before run following command. To run on institutional clusters, see installation notes for individual packages on the module list of the cluster. Examples of installation on Frontera are below.
 
-# Installing on institutional clusters:
```
carlosgjs, line 43: This looks like the same installation steps as the single node installation, so maybe it't not needed?

mdenolle: it looks like we wrote this in case some of the packages are already installed in the default environment of some HPC cluster.  PHDF5 is a parallel HDF5 library (https://support.hdfgroup.org/HDF5/PHDF5/). What would you recommend in the case where we would need phdf5 instead pf hdf5 libraries?

carlosgjs: @mdenolle got it. I missed the first line and the fact that `h5py` is omitted from the module list. , but if someone follows the standard instructions and runs the normal `pip install -r requirements.txt`, would this not work too? Installing hp5y in the virtual environment is not much overhead and we can keep things simple with a single set of instructions.

That said, I'm not familiar with Frontera/phdf5 so if this doesn't work I'm happy to revert. Do you have an environment where this could be tested?

mdenolle: OK I asked a colleague at TACC and it still requires h5py so we can. probably ignore that section

niyiyu (approved):
All looks good to me!

README.md
```diff
@@ -21,34 +21,31 @@ Jiang, C. and Denolle, M. "NoisePy: a new high-performance python tool for seism
...
+The nature of NoisePy being composed of python scripts allows flexible package installation, which is essentially to build dependent libraries the scripts and related functions live upon. We recommend using [conda](https://docs.conda.io/en/latest/) or [pip](https://pypi.org/project/pip/) to install the library due to their convenience. Below are command lines we have tested to create a python environment to run NoisePy. Note that the test is performed on `macOS Mojave (10.14.5)`, so it could be slightly different for other OS. 
+
+## Pre-requisites
+
+An MPI installation is required. E.g. for macOS using (brew)[https://brew.sh/] :
+```sh
+brew install open-mpi
+```
```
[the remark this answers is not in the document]
carlosgjs: I'll add it as an alternative path, since we are giving instructions for both conda and pip.

PR #58, opened 2023-03-21
Add stacking to noisepy.py and as part of merge tests

carlosgjs:
As noted in the README, this PR adds the `stack` cmd:
```
usage: noisepy.py stack [-h] [--path PATH] --method
                        {linear,pws,robust,nroot,selective,auto_covariance,all}

optional arguments:
  --path PATH           Directory to look for input files (default:
                        /Users/carlosg/Documents/SCAL)
  --method {linear,pws,robust,nroot,selective,auto_covariance,all}
                        Stacking method (default: None)

```
It also adds this step into the verification tests.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -7,7 +7,7 @@ on:
 
 env:
   python_version: 3.8
-  testdir: ./test
+  testdir: ~/test
 jobs:
   s0_download:
     runs-on: ubuntu-22.04
@@ -20,11 +20,11 @@ jobs:
         python-version: ${{env.python_version}}
     - name: Test Download (S0)
       working-directory: ./src
-      run: python noisepy.py download --start 2019_02_01_00_00_00 --end 2019_02_01_02_00_00 --stations ARV,BAK --inc_hours 1 --path ../$testdir
+      run: python noisepy.py download --start 2019_02_01_00_00_00 --end 2019_02_01_02_00_00 --stations ARV,BAK --inc_hours 1 --path $testdir
     - name: Cache data
       uses: actions/cache@v3
       with:
-        key: download_data
+        key: download_data_2
         path: ${{env.testdir}}
   s1_cross_correlate:
     strategy:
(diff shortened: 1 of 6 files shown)


src/S2_stacking.py
```diff
@@ -37,339 +35,345 @@
 #########PARAMETER SECTION##############
 ########################################
```
carlosgjs, line 37: No functional changes, just wrapping the script in a function

.github/workflows/test.yaml
```diff
@@ -29,8 +29,14 @@ jobs:
...
+      run: python noisepy.py download --start 2019_02_01_00_00_00 --end 2019_02_01_02_00_00 --stations ARV,BAK --inc_hours 1 --path .
     - name: Test Cross-Correlation (S1)
       working-directory: ./src
       run: python noisepy.py cross_correlate --path .
+    - name: Test Stacking (S2)  
+      working-directory: ./src
+      run: |
+        python noisepy.py stack --path . --method linear
+        rm -rf STACK
+        python noisepy.py stack --path . --method pws
```
carlosgjs, line 41: Should other methods be tested here? E.g. `all`?

niyiyu (approved):

src/S2_stacking.py
```diff
@@ -37,339 +35,345 @@
...
 # define new stacking para
 keep_substack= False                                                # keep all sub-stacks in final ASDF file
 flag         = True                                                 # output intermediate args for debugging
-stack_method = 'linear'                                             # linear, pws, robust, nroot, selective, auto_covariance or all
 
 # new rotation para
 rotation     = True                                                 # rotation from E-N-Z to R-T-Z 
 correction   = False                                                # angle correction due to mis-orientation
-if rotation and correction:
-    corrfile = os.path.join(rootpath,'meso_angles.txt')             # csv file containing angle info to be corrected
```
mdenolle, line 57: I see a hardcoded file here that was quite specific to a project we had. Could we write this filename as a separate variable input. it matters a lot for borehole and offshore stations where there is less control on how the sensor is oriented. For a later modification: we should also have this file input as a CSV and use dataframes instead for metadata.

mdenolle (approved):

src/S2_stacking.py
carlosgjs: Got it. I've created an issue for this task, so that this PR maintaince its single focus.

PR #64, opened 2023-03-27
Enabe isort pre-commit check

carlosgjs:


niyiyu (approved):
Looks like most change are re-sort the packages when being imported.

carlosgjs:
> Looks like most change are re-sort the packages when being imported. Thanks for bringing this! Is that a one-time change? Or it automatically update all when new files are committed?

@niyiyu The pre-commit hook will auto-reformat before all commits. You need to install it by running:
```
pip install pre-commit>=3.2.0
pre-commit install
```
See: https://pre-commit.com/

PR #66, opened 2023-03-28
Nick/precommit

nrich20:
Enabled the black pre-commit function. Ran it through all the files and it successfully catches and modifies files to comply with the black coding standard.


carlosgjs (approved):
Nice!

PR #68, opened 2023-03-29
Run unit tests as part of CI

carlosgjs:
- Moved unit tests (`test_whiten.py` and `test_stretching.py`) to a new `unittests` directory to separate unit tests and one-off/manual testing scripts
- Refactor `test_whiten` to display plots only when run manually and not during unit tests (which blocks until the plot is closed)
- Run unit tests as part of CI


niyiyu (approved):

niyiyu:
All looks good!

PR #70, opened 2023-03-31
Add release workflow for container

carlosgjs:
This PR adds a workflow that builds and pushes a noisepy container when a release is created. Here's the package from a test run: https://github.com/mdenolle?tab=packages.

The container can be run as follows (e.g. for the `cross_correlate` step, assuming data is in the local `~/tmp` directory):
```
docker run -v ~/tmp:/tmp ghcr.io/mdenolle/noisepy:latest cross_correlate --path /tmp
```

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+*   @mdenolle @niyiyu @carlosgjs
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,1 +1,1 @@
-*   @mdenolle @niyiyu @carlosgjs
+*   @mdenolle @niyiyu @carlosgjs @chengxinjiang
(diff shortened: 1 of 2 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+*   @mdenolle @niyiyu @carlosgjs
```
mdenolle, line 1: could please add @chengxinjiang who was also a core developer?

mdenolle (changes requested):

.github/CODEOWNERS
carlosgjs: Sure, done! FYI, the CODEOWNERS listed here will be added as reviewers by default for any PRs. We can customize as needed.

niyiyu (approved):
All looks good!

mdenolle (approved):
looks good!

PR #73, opened 2023-04-04
Remove uses of sys.exit()

carlosgjs:
Explicit calls to sys.exit() make the software hard to use as a library (vs standalone scripts). They are also unnecessary in most cases. The calls fall into two categories:

- `sys.exit()` calls for rank 0 nodes in mpi. These are just removed since the process will naturally exit anyway.
- `sys.exit()` calls after hitting an error. In this case it's better to raise an exception. This allows the caller to catch/handle it if needed and if not caught the process will exit with an error code anyway.

### DIFF
--- a/src/S0A_download_ASDF_MPI.py
+++ b/src/S0A_download_ASDF_MPI.py
@@ -185,8 +185,9 @@ def download(
                             level="response",
                         )
                     except Exception as e:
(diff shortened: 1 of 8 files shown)


mdenolle (approved):
looks good!

PR #76, opened 2023-04-06
Making noisepy a PyPi package

carlosgjs:
Summary of changes:

- Move dependencies and settings to `pyproject.toml`
- Update module imports to be relative to the package
- Update integration tests to use the installed `noisepy` script

Verified package can be successfully published: https://pypi.org/project/noisepy-seis/  and that `pip install noisepy-seis` works.

TODO: 
- Build and publish the package as part of the release workflow


src/noisepy/seis/__init__.py
```diff
@@ -0,0 +1,9 @@
+# no-qa F401
+from .S0A_download_ASDF_MPI import download  # noqa: F401
+from .S1_fft_cc_MPI import cross_correlate  # noqa: F401
+from .S2_stacking import stack  # noqa: F401
+
+__version__ = "0.5.4"
```
lsetiawan, line 6: This manual package versioning is probably fine for now, but eventually it would be better to have auto versioning based on the version control system. This way, when you make a release it will just grab the version from the release tag. Now that you're using hatch, I suggest adding `hatch-vcs` https://github.com/ofek/hatch-vcs

pyproject.toml
```diff
@@ -0,0 +1,77 @@
...
+
+[tool.hatch.build.targets.wheel]
+packages = ["src/noisepy"]
+
+[project.optional-dependencies]
+dev = [
+    "pytest>=7.2.2",
+    "memory-profiler>=0.61",
+    "pre-commit>=3.2",
+    "hatchling>=1.14"
```
lsetiawan, line 66: This is already required by the build system, shouldn't need to list this out for dev

.gitignore
```diff
@@ -1 +1,2 @@
 **/__pycache__
+/dist
```
lsetiawan, line 2: Maybe fill this file with the default github's python .gitignore: https://github.com/github/gitignore/blob/main/Python.gitignore

unittests/test_stretching.py
```diff
@@ -4,7 +4,7 @@
 import pytest
 from obspy.signal.invsim import cosine_taper
 
-from noise_module import stretching, stretching_vect
+from noisepy.seis.noise_module import stretching, stretching_vect
```
lsetiawan, line 7: This is great!

Question: Is there a specific reason that there are 2 test folders: test and unittests?

.pre-commit-config.yaml
```diff
@@ -18,14 +18,11 @@ repos:
     rev: 5.12.0
     hooks:
       - id: isort
```
carlosgjs, line 20: moved to `pyproject.toml`

src/noisepy/seis/__init__.py
carlosgjs: Gotcha, I'll do that in a subsequent PR

unittests/test_stretching.py
carlosgjs: Yeah, legacy. The `test` directory used to contain both one-of manual testing scripts and unit tests. I split the unit tests into their own directory to make it easier to run them via pytest during CI. Perhaps I'll rename:
`tests` --> `scripts`
`unittests` --> `tests`
wdyt?

pyproject.toml
```diff
@@ -0,0 +1,76 @@
...
+]
+dependencies = [
+    "h5py>=3.8.0",
+    "mpi4py>=3.1.4",
+    "numba>=0.56.4",
+    "numpy>=1.22.0",
+    "pandas>=1.5.3",
+    "pyasdf>=0.7.5",
+    "pycwt>=0.3.0a22",
+    "wheel>=0.38.4",
```
lsetiawan, line 43: No need for wheel either since it's only required during build.

unittests/test_stretching.py
lsetiawan: I like that better! It's definitely a lot more clear.

lsetiawan (approved):
The build components looks great. I think having the `seis` as a subpackage structure makes it a lot more clear and pythonic. Nice work @carlosgjs! 😄 

I just tried locally to install the package with the current configuration and it worked. I also tried to build the package to a wheel and distribution with python `build` (see https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives) and that also worked! ~~Though I didn't try installing the package from the wheel or distribution.~~

**UPDATE**: Just tried installing from wheel just now and that worked. Also, the tarball looks good 😸

lsetiawan:
Another comment I have not related to this PR is that the repo is so big to download!

.gitignore
carlosgjs: Created issue #77

PR #79, opened 2023-04-07
Add pypi github workflow

lsetiawan:
## Overview

This PR adds github workflow to build and push packages to both test pypi and main pypi during releases and push to master branch.

### DIFF
--- a/.github/workflows/pypi.yaml
+++ b/.github/workflows/pypi.yaml
@@ -1,0 +1,99 @@
+# Publish archives to PyPI and TestPyPI using GitHub Actions
+name: Publish to PyPI
+
+# Only run for tagged releases
+on:
+  release:
+    types:
+      - released
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+jobs:
+  build-artifact:
+    name: Build noisepy-seis package
+    runs-on: ubuntu-20.04
+    if: github.repository == 'mdenolle/NoisePy'
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: Set up Python
+      uses: actions/setup-python@v4.5.0
+      with:
+        python-version: 3.9
+
+    - name: Install dependencies
+      run: python -m pip install build
+
+    - name: Build source and wheel distributions
+      run: |
+        python -m build
+        echo ""
+        echo "Generated files:"
+        ls -lh dist/
+    - uses: actions/upload-artifact@v3
+      with:
+        name: releases
+        path: dist
+
+  test-built-dist:
+    name: Test noisepy-seis package
+    needs: build-artifact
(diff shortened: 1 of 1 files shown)


lsetiawan:
@carlosgjs you'd need to set up the `secrets.TEST_PYPI_API_TOKEN` and `secrets.PYPI_API_TOKEN` for this to work.

.github/workflows/pypi.yaml
```diff
@@ -0,0 +1,99 @@
...
+        python -m build
+        echo ""
+        echo "Generated files:"
+        ls -lh dist/
+    - uses: actions/upload-artifact@v3
+      with:
+        name: releases
+        path: dist
+
+  test-built-dist:
```
carlosgjs, line 42: Is it overkill to upload to the test index and the download it back down? It adds complexity and maintenance cost (e.g. needing a separate token). Can we test the local wheel instead? E.g.
```
pip install dist/noisepy_seis-*.whl
python -c "import noisepy.seis; print(noisepy.seis.__version__)"
...
```
That way we are testing the contents of the package (what we are building) and not the PyPi service.

lsetiawan: It's usually good to test the publishing and ensuring that the package works against the test pypi. This is to ensure that the behavior will work when pushing to the "production" pypi. Additionally this can provide a way to distribute the package with internal users for testing before officially releasing it to the public. This is really the purpose of the test pypi.

lsetiawan: In regards of the simple wheel test is to ensure that it works since when you put into conda, it will perform those exact checks. Better to do that during testing then after the package is published :)

lsetiawan:
This is a duplicate of #80, but have some differences. Sorry @carlosgjs! Didn't mean to step on your work, just wanted to share my experience with packaging and distribution, preparing for both pypi and conda.

.github/workflows/pypi.yaml
```diff
@@ -0,0 +1,99 @@
...
+  push:
+    branches:
+      - master
+  workflow_dispatch:
+
+jobs:
+  build-artifact:
+    name: Build noisepy-seis package
+    runs-on: ubuntu-20.04
+    if: github.repository == 'mdenolle/NoisePy'
```
carlosgjs, line 18: Why is this needed?

carlosgjs (approved):

.github/workflows/pypi.yaml
lsetiawan: I do this check, to make sure that this only runs on this repo and not anyone else's fork. This again is probably an overkill since pypi doesn't allow duplicate packages.

lsetiawan:
Closing in favor of #80

PR #80, opened 2023-04-07
Add pypi publishing to the release workflow

carlosgjs:
- Adds building/testing/publishing jobs from PR #79 
- Use ubuntu 20.04 and python 3.8
- Use published noisepy package to build docker container
- Pass the published version number of the docker build

Latest release in PyPI:

https://pypi.org/project/noisepy-seis/0.5.9/

Latest docker: https://github.com/mdenolle/NoisePy/pkgs/container/noisepy/84301205?tag=v0.5.9


Dockerfile
```diff
@@ -6,10 +6,6 @@ RUN apt-get update && \
     apt-get install -y python3-pip && \
     pip3 install --upgrade pip
 
-# Copy and install requirements first so this layer doesn't change when copy the local files below
-COPY requirements.txt .
-RUN pip3 install -r requirements.txt
-ADD . .
-WORKDIR /src
-ENTRYPOINT ["python3", "noisepy.py"]
+RUN pip3 install noisepy-seis
```
lsetiawan, line 9: Probably good to add an `ARG` at the very top and put in specific version here, that way when you're tagging the docker image, you know the exact version of the package that you're using? See https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact for docs.

Example:

```
ARG VERSION=0.5.4
...

RUN pip3 install noisepy-seis==${VERSION}
```

What do you think?

uwcdc (commented):
I agree with @lsetiawan adding a VC for the docker image will be helpful

Dockerfile
carlosgjs: Makes sense!

Dockerfile
```diff
@@ -1,15 +1,15 @@
...
     apt-get install -y python3-pip && \
     pip3 install --upgrade pip
 
-# Copy and install requirements first so this layer doesn't change when copy the local files below
-COPY requirements.txt .
-RUN pip3 install -r requirements.txt
-ADD . .
-WORKDIR /src
-ENTRYPOINT ["python3", "noisepy.py"]
+ARG VERSION
```
lsetiawan, line 12: `ARG` should be at the very top of the file before `FROM`.

carlosgjs: The problem with having it higher is that it invalidates the layer caching for any RUN commands below it: https://docs.docker.com/engine/reference/builder/#impact-on-build-caching

lsetiawan (approved):

PR #83, opened 2023-04-13
RFC: Data types and abstractions

carlosgjs:
This PR is an RFC (request for comments) to the introduction of certain abstractions. After digging a bit more deeply into the code (specifically S1) I noticed that multiple concerns are interleaved:

- The way the data is stored (h5 vs ms, StationXML?)
- Parallelization through MPI
- The core cross-correlation computation

In addition, there are no custom types to capture the data dependencies, so most of the code uses basic data types in a loose way. E.g. flat lists of numbers (`lat, lon`), and strings (`sta_list, all_tags`). The fact that a `lat, lon` are properties of a *station* and that a station has multiple *channels* is all implicit in the implementation.

This PR proposes a few changes:
- Introduction of semantic types, E.g. `Channel` and `Station` for better readability and consistency.
- Use of abstractions for the data input/output so that the core computation can be agnostic to the way data is stored.
- Isolating the parallelization mechanism as much as possible. Ie. I should be able to do core computations without having MPI installed if I'm not using MPI. This also makes the main algorithms easier to read.

### DIFF
--- a/src/noisepy/seis/datatypes.py
+++ b/src/noisepy/seis/datatypes.py
@@ -1,0 +1,161 @@
+from abc import ABC, abstractmethod
+from dataclasses import dataclass
+from datetime import datetime
+from enum import Enum
+from typing import List
+
+import numpy as np
+from datetimerange import DateTimeRange
+
+from noisepy.seis import noise_module
+
+
+@dataclass
+class ChannelType:
+    """
+    A type of channel, e.g. BHN, but not associated with a particular station
+    """
+
+    name: str
+
+    def __init__(self, name):
+        assert len(name) == 3, "A channel type name should be length 3"
+        self.name = name
+
+    def get_orientation(self) -> str:
+        return self.name[-1]
+
+
+@dataclass
+class Station:
+    network: str  # 2 chars
+    name: str  # 3-5 chars
+    lat: float
+    lon: float
+    elevation: float
+    location: str
+
+    def __repr__(self) -> str:
+        return f"{self.network}.{self.name}"
+
+
+class CorrelationMethod(Enum):
+    XCORR = 1
+    DECONV = 2
+
+
+@dataclass
+class CrossCorrelationParameters:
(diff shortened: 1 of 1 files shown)


src/noisepy/seis/datatypes.py
```diff
@@ -0,0 +1,161 @@
...
+        parameters: CrossCorrelationParameters,
+        data: np.ndarray,
+    ):
+        pass
+
+    @abstractmethod
+    def read(self, chan1: Channel, chan2: Channel, start: datetime, end: datetime) -> np.ndarray:
+        pass
+
+
```
carlosgjs, line 120: Below is a pseudo-implementation to illustrate how the core code might change as a result

src/noisepy/seis/datatypes.py
```diff
@@ -0,0 +1,161 @@
...
+    @abstractmethod
+    def get_channels(self) -> List[Channel]:
+        pass
+
+    @abstractmethod
+    def get_timespans(self) -> List[DateTimeRange]:
+        pass
+
+
```
carlosgjs, line 88: We can have multiple implementations of this interface. E.g. `H5DataSource`, `MiniSeedDataSource` or `S3ZarrDataSource`.

src/noisepy/seis/datatypes.py
```diff
@@ -0,0 +1,161 @@
...
+    @abstractmethod
+    def read(self, chan1: Channel, chan2: Channel, start: datetime, end: datetime) -> np.ndarray:
+        pass
+
+
+# dummy
+def fft(d):
+    pass
+
+
```
carlosgjs, line 125: The main code can be greatly simplified by abstracting out the data I/O and using higher level types (of course the main computation is stubbed out here)

mdenolle (approved):
looks like a good pseudo-implementation!

carlosgjs:
replaced by PR #85

PR #85, opened 2023-04-14
Base data types and ASDFStore implementation

carlosgjs:
This PR replaces  https://github.com/mdenolle/NoisePy/pull/83 and adds a concrete implementation of an ASDFDataStore. 

# Context
After digging a bit more deeply into the code (specifically S1) I noticed that multiple concerns are interleaved:

- The way the data is stored (h5 vs ms, StationXML?). E.g. there are 8 checks like [`if format == 'h5'`](https://github.com/mdenolle/NoisePy/blob/21322356b8d2fb0ebbcc32632bd1c5ac1a156dbf/src/noisepy/seis/S1_fft_cc_MPI.py#L125).
- Parallelization through MPI
- The core cross-correlation computation

In addition, there are no custom types to capture the data dependencies, so most of the code uses basic data types in a loose way. E.g. flat lists of numbers (lat, lon), and strings (sta_list, all_tags). The fact that `lat` and `lon` are properties of a `station` and that a `station` has multiple `channels` is all implicit in the S1 implementation.

# Proposed changes:

- Introduction of semantic types, E.g. Channel and Station for better readability and consistency.
- Use of abstractions for the data input/output so that the core computation can be agnostic to the way data is stored.
- Initial implementation of the concrete ASDFDataStore

# Suggested reading order:

- Inline comments in https://github.com/mdenolle/NoisePy/pull/83
- datatypes.py
- stores.py
- asdfstore.py
- test_*.py


mdenolle (approved):
these new data types and store look good

PR #94, opened 2023-05-02
SCEDC bug fixes, channel filtering

carlosgjs:
Feature: Allow an optional argument to the `cross_correlate` function to limit the channels used.

Bug fixes:
- Fix file name from channel name 
- Cache the inventory in memory, since reading the disk cache is still expensive
- Use default config if now `download.txt` is available

### DIFF
--- a/src/noisepy/seis/S1_fft_cc_MPI.py
+++ b/src/noisepy/seis/S1_fft_cc_MPI.py
@@ -62,8 +62,8 @@ def cross_correlate(
                 raw_store: Store to load data from
                 fft_params: Parameters for the FFT calculations
                 cc_store: Store for saving cross correlations
-                channel_filter: Function to decide whether a station should be used or not,
-                                if None, all stations are used
+                channel_filter: Function to decide whether a channel should be used or not,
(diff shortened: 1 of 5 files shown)


src/noisepy/seis/noise_module.py
```diff
@@ -523,7 +523,7 @@ def cut_trace_make_stat(fc_para: ConfigParameters, ch_data: ChannelData):
     # trace_madS = np.zeros(nseg,dtype=np.float32)
     trace_stdS = np.zeros(nseg, dtype=np.float32)
     dataS = np.zeros(shape=(int(nseg), int(npts)), dtype=np.float32)
-    dataS_t = np.zeros(nseg, dtype=np.float)
+    dataS_t = np.zeros(nseg, dtype=np.float32)
```
carlosgjs, line 526: @mdenolle `np.float` has been deprecated. I changed this to `np.float32` to be consistent with the other arrays used here. But let me know if this particular one should be `np.float64`.

mdenolle (approved):
looks great! thank you for the improvements

PR #96, opened 2023-05-03
Run preprocessing during S1

carlosgjs:
Changes: 
 - Use `ConfigParameters` in S0
- Explicitly pass directory names to the various steps, instead of assuming names such as `RAW_DATA` given that other data sources don't use this convention (e.g. SCEDC S3 bucket) 
- Use well defined types for `preprocess_raw`
- Add `respdir` and `rm_resp_out` to ConfigParameters
- Read the channel information and filter to channels having the closest `sampling_rate>=config.freq_samp`
- Add support in `RawDataStore` and `ChannelCatalog` to get an `Inventory` object (to be passed into `preprocess_raw`)
- Call `preprocess_raw` on the data during S1


src/noisepy/seis/noise_module.py
```diff
@@ -263,8 +266,8 @@ def preprocess_raw(st, inv, prepro_para, date_info):
                     print("removing response for %s using inv" % st[0])
                     st[0].attach_response(inv)
                     st[0].remove_response(output=rm_resp_out, pre_filt=pre_filt, water_level=60)
-                except Exception:
-                    print("WARNING: Failed to remove response from %s. Returning empty stream." % st[0])
+                except Exception as e:
+                    print("WARNING: Failed to remove response from %s. Returning empty stream. %s" % (st[0], e))
```
carlosgjs, line 270: @mdenolle if I test with `rmresp="inv"`, using the `SCEDCS3DataStore` (S3 data), this fails with the error:
```
Can not use evalresp on response with no response stages.
```
In this case the inventory is obtained using:
```
         client = Client(self.url_key)
         inventory = client.get_stations_bulk(bulk_station_request, level="channel")
```
Maybe this is not a valid configuration to use? I think this should be using the files from: s3://scedc-pds/FDSNstationXML/? These are all .XML files, e.g. (`CI_ABL.XML`). Is this what I should use? And which type of `rm_resp` does this correspond to? (`spectrum|RESP|polozeros`)

carlosgjs: This has been resolved in https://github.com/mdenolle/NoisePy/pull/97

mdenolle (approved):
excellent!

PR #97, opened 2023-05-04
XMLStationChannelCatalog implementation

carlosgjs:
- ChannelCatalog implementation that reads XML files obtained from `s3://scedc-pds/FDSNstationXML` for inventory information.
- Moved the channel filtering from S1 into the `SCEDCS3DataStore` to avoid fully loading a large number of channels (>4000), when only a subset (~10) is used
- Updates to the CLI (`main.py`) to use the new channel catalog and take the list of stations to filter on for the `cross_correlation` step.


mdenolle (approved):
superb, thank you for taking the initiative for this

PR #98, opened 2023-05-04
Generate version from git tags

carlosgjs:
Using [hatch-vcs](https://github.com/ofek/hatch-vcs)

Test run: https://github.com/mdenolle/NoisePy/actions/runs/4887451147

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -17,7 +17,10 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v3
-
+    - name: Get history and tags for SCM versioning to work
+      run: |
+        git fetch --prune --unshallow
+        git fetch --depth=1 origin +refs/tags/*:refs/tags/*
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yaml
```diff
@@ -17,7 +17,10 @@ jobs:
     steps:
     - name: Checkout
       uses: actions/checkout@v3
-
+    - name: Get history and tags for SCM versioning to work
+      run: |
+        git fetch --prune --unshallow
+        git fetch --depth=1 origin +refs/tags/*:refs/tags/*
```
lsetiawan, line 23: No need for this step. The `actions/checkout` has an attribute that will pull this. Like so:
```
- name: Checkout repo
  uses: actions/checkout@v3
  with:
    fetch-depth: 0 # Fetch all history for all branches and tags.
```

lsetiawan (approved):
All looks great other than removing the things with fetching the tags and branches.

PR #101, opened 2023-05-04
Minor renaming of files , readme

mdenolle:
minor text change


README.md
```diff
@@ -71,9 +71,58 @@ Seats, K. J., Jesse F. L., and German A. P. "Improved ambient noise correlation
 *Jiang, C. and Denolle, M. "NoisePy: a new high-performance python tool for seismic ambient noise seismology." _Seismological Research Letter_ 91, no. 3 (2020): 1853–1866..
 ** Yuan, C., Bryan, J. T., and Denolle, M. "Numerical comparison of time-, frequency- and wavelet-domain methods for coda wave interferometry." _Geophysical Journal International_ 226, no. 2 (2021): 828-846.
 
+## Parameters to configure
```
carlosgjs, line 74: @mdenolle I'd prefer not to duplicate this here ([DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)), since it could be error prone if the code and the README diverge. An alternative is to put a link here to `blob/master/src/noisepy/seis/datatypes.py`. This will point people to the source of truth for the parameters and the default values. We can even move the `ConfigParameters` class to its own file so that it's easier to link directly to it.

src/noisepy/seis/plotting_modules.py
```diff
@@ -221,23 +225,23 @@ def plot_substack_cc(sfile, freqmin, freqmax, disp_lag=None, savefig=True, sdir=
...
-            ax1.set_xticks(t)
+            # ax1 = fig.add_subplot(413)
+            # ax1.set_title("stacked and filtered at %4.2f-%4.2f Hz" % (freqmin, freqmax))
+            # ax1.plot(
+            # np.arange(-disp_lag, disp_lag + dt, dt),
+            # np.mean(data, axis=0),
+            # "k-",
+            # linewidth=1,
+            # )
+            # ax1.set_xticks(t)
```
carlosgjs, line 236: Better to remove unused code
```suggestion
```

carlosgjs:
@mdenolle can you change the title of the PR to something more descriptive? It will become part of the commit history when the PR is merged. Also, a description of the changes would be great. thx!

README.md
mdenolle: Interesting. I think the configuration parameters are the most important information for the users and should be highly visible somewhere. You are right that a readme is probably not appropriate, should we consider in the docs/examples/tutorial.rst or docs/examples/example.rst?

carlosgjs: Yeah, I think the docs are a better place for this. I see there's already some code in the repo for building them with sphinx. I think the best path is to connect this to readthedocs.io so that the docs are built and published on every PR. This will generate a nice documentation page right from the source code, hence removing the need for duplication. It would look something like this: https://example-python.readthedocs.io/en/latest/example.html.

I'm ok with placing the parameters temporarily somewhere else and then moving to readthedocs later on...

README.md
```diff
@@ -71,9 +71,58 @@ Seats, K. J., Jesse F. L., and German A. P. "Improved ambient noise correlation
...
+
+The parameters of the workflow are saved into an object called ``ConfigParameter``. The parameters
+
+ *   dt: float = 0.05  # TODO: dt should be 1/sampling rate
```
carlosgjs, line 78: If we want to add this here temporarily we should remove the TODOs

carlosgjs (approved):

PR #102, opened 2023-05-05
Support reading directly from S3 buckets

carlosgjs:
- Replace direct filesystem access with `fsspec` to support both S3 and local files
- Update unit tests to verify both local data and S3
- Add debug configuration for S3 case in `launch.json`

Test run:
```
$ noisepy cross_correlate --raw_data_path s3://scedc-pds/continuous_waveforms/2002/2002_002/ --stations "SBC,RIO,DEV,HEC,RLR,SVD,RPV,BAK" --xml_path "s3://scedc-pds/FDSNstationXML/CI/" --freq_norm rma 

2023-05-05 11:04:52,548 INFO scedc_s3store __init__ Init: 1 timespans and 42 channels
2023-05-05 11:05:11,349 INFO S1_fft_cc_MPI _filter_channel_data Picked 20.0 as the closest sampling frequence to 20. Filtered to 21/42 channels
unreadable garbarge 9
it takes  75.59s to process the chunk of 2002-01-02T00:00:00+0000 - 2002-01-03T00:00:00+0000
it takes  75.59s to process step 1 in total
```

### DIFF
--- a/src/noisepy/seis/channelcatalog.py
+++ b/src/noisepy/seis/channelcatalog.py
@@ -11,6 +11,7 @@ from obspy import UTCDateTime, read_inventory
 from obspy.clients.fdsn import Client
 
 from .datatypes import Channel, Station
+from .utils import get_filesystem
 
 logger = logging.getLogger(__name__)
 
@@ -57,22 +58,24 @@ class XMLStationChannelCatalog(ChannelCatalog):
     A channel catalog that reads <station>.XML files from a directory
     """
 
-    def __init__(self, xmldir: str) -> None:
+    def __init__(self, xmlpath: str) -> None:
         super().__init__()
-        self.xmldir = xmldir
-        if not os.path.exists(self.xmldir):
-            raise Exception(f"The XML Station file directory '{xmldir}' doesn't exist")
+        self.xmlpath = xmlpath
+        self.fs = get_filesystem(xmlpath)
+        if not self.fs.exists(self.xmlpath):
+            raise Exception(f"The XML Station file directory '{xmlpath}' doesn't exist")
 
     def get_inventory(self, timespan: DateTimeRange, station: Station) -> obspy.Inventory:
-        xmlfile = os.path.join(self.xmldir, f"{station.network}_{station.name}.xml")
+        xmlfile = os.path.join(self.xmlpath, f"{station.network}_{station.name}.xml")
         return self._get_inventory_from_file(xmlfile)
 
     @lru_cache
     def _get_inventory_from_file(self, xmlfile):
-        if not os.path.exists(xmlfile):
+        if not self.fs.exists(xmlfile):
             logger.warning(f"Could not find StationXML file {xmlfile}. Returning empty Inventory()")
             return obspy.Inventory()
-        return read_inventory(xmlfile)
+        with self.fs.open(xmlfile) as f:
+            return read_inventory(f)
 
 
 class FDSNChannelCatalog(ChannelCatalog):
--- a/src/noisepy/seis/channelcatalog.py
+++ b/src/noisepy/seis/channelcatalog.py
@@ -55,7 +55,7 @@ class ChannelCatalog(ABC):
 
 class XMLStationChannelCatalog(ChannelCatalog):
     """
-    A channel catalog that reads <station>.XML files from a directory
+    A channel catalog that reads <station>.XML files from a directory or an s3://... bucket path.
     """
 
     def __init__(self, xmlpath: str) -> None:
--- a/src/noisepy/seis/channelcatalog.py
+++ b/src/noisepy/seis/channelcatalog.py
@@ -1,5 +1,4 @@
 import logging
-import os
 from abc import ABC, abstractmethod
 from functools import lru_cache
 
@@ -11,7 +10,7 @@ from obspy import UTCDateTime, read_inventory
 from obspy.clients.fdsn import Client
 
 from .datatypes import Channel, Station
-from .utils import get_filesystem
+from .utils import fs_join, get_filesystem
 
 logger = logging.getLogger(__name__)
 
@@ -66,7 +65,7 @@ class XMLStationChannelCatalog(ChannelCatalog):
             raise Exception(f"The XML Station file directory '{xmlpath}' doesn't exist")
 
     def get_inventory(self, timespan: DateTimeRange, station: Station) -> obspy.Inventory:
-        xmlfile = os.path.join(self.xmlpath, f"{station.network}_{station.name}.xml")
+        xmlfile = fs_join(self.xmlpath, f"{station.network}_{station.name}.xml")
         return self._get_inventory_from_file(xmlfile)
 
     @lru_cache
--- a/src/noisepy/seis/channelcatalog.py
+++ b/src/noisepy/seis/channelcatalog.py
@@ -54,7 +54,7 @@ class ChannelCatalog(ABC):
 
 class XMLStationChannelCatalog(ChannelCatalog):
     """
-    A channel catalog that reads <station>.XML files from a directory or an s3://... bucket path.
+    A channel catalog that reads <station>.XML files from a directory or an s3://... bucket url path.
     """
 
     def __init__(self, xmlpath: str) -> None:
--- a/src/noisepy/seis/main.py
+++ b/src/noisepy/seis/main.py
@@ -14,7 +14,7 @@ from .S0A_download_ASDF_MPI import download
 from .S1_fft_cc_MPI import cross_correlate
 from .S2_stacking import stack
 from .scedc_s3store import SCEDCS3DataStore
-from .utils import get_filesystem
+from .utils import fs_join, get_filesystem
 
 # Utility running the different steps from the command line. Defines the arguments for each step
 
@@ -37,7 +37,7 @@ def valid_date(d: str) -> str:
 
 def initialize_fft_params(raw_dir: str) -> ConfigParameters:
     params = ConfigParameters()
-    dfile = os.path.join(raw_dir, "download_info.txt")
+    dfile = fs_join(raw_dir, "download_info.txt")
     if os.path.isfile(dfile):
         down_info = eval(open(dfile).read())  # TODO: do proper json/yaml serialization
         params.samp_freq = down_info["samp_freq"]
@@ -62,7 +62,7 @@ def create_raw_store(raw_dir: str, sta_list: List[str], xml_path: str):
     fs = get_filesystem(raw_dir)
 
     def count(pat):
-        return len(fs.glob(os.path.join(raw_dir, pat)))
+        return len(fs.glob(fs_join(raw_dir, pat)))
 
     # Use heuristics around which store to use by the files present
     if count("*.h5") > 0:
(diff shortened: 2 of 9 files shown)


src/noisepy/seis/channelcatalog.py
```diff
@@ -54,25 +55,27 @@ def get_inventory(self, timespan: DateTimeRange, station: Station) -> obspy.Inve
...
-        if not os.path.exists(self.xmldir):
-            raise Exception(f"The XML Station file directory '{xmldir}' doesn't exist")
+        self.xmlpath = xmlpath
+        self.fs = get_filesystem(xmlpath)
+        if not self.fs.exists(self.xmlpath):
+            raise Exception(f"The XML Station file directory '{xmlpath}' doesn't exist")
 
     def get_inventory(self, timespan: DateTimeRange, station: Station) -> obspy.Inventory:
-        xmlfile = os.path.join(self.xmldir, f"{station.network}_{station.name}.xml")
+        xmlfile = os.path.join(self.xmlpath, f"{station.network}_{station.name}.xml")
```
lsetiawan, line 69: This assumes that `xmlpath` is a directory path. Maybe in `__init__` you there need to be a check for this also, otherwise, you might get something like `s3://mybucket/path/myxml.xml` which would cause this join to give you the wrong path to actual `xml` file.

src/noisepy/seis/scedc_s3store.py
```diff
@@ -72,12 +74,13 @@ def read_data(self, timespan: DateTimeRange, chan: Channel) -> ChannelData:
             f"{chan.station.network}{chan.station.name.ljust(5, '_')}{chan.type.name}"
             f"{chan.station.location.ljust(3, '_')}"
         )
-        filename = os.path.join(self.directory, f"{chan_str}{timespan.start_datetime.strftime('%Y%j')}.ms")
-        if not os.path.exists(filename):
+        filename = os.path.join(self.path, f"{chan_str}{timespan.start_datetime.strftime('%Y%j')}.ms")
```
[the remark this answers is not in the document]
lsetiawan: This problem is such a mess. I've had to deal with this a lot for echopype development. A utility module that sanitizes or checks for this would be best. As an example, we have done this (though we haven't quite cleaned this up 😬 ): https://github.com/OSOceanAcoustics/echopype/blob/43a66bb4fca880e4ae52176c14b6a5f1781bcab1/echopype/utils/io.py#L236-L240. Additionally for anything with URL, you can do a check for `://` in the string since all url is built that way, and force the path join to be with `/` by doing the python string join method.

src/noisepy/seis/utils.py
```diff
@@ -0,0 +1,7 @@
+import fsspec
+
+S3PREFIX = "s3"
+
+
+def get_filesystem(path: str) -> fsspec.AbstractFileSystem:
+    return fsspec.filesystem(S3PREFIX, anon=True) if path.startswith(S3PREFIX) else fsspec.filesystem("file")
```
lsetiawan, line 7: This check is not really robust since what if a path is like `s3/mydir/test`? Can't really confirm if it's a url or not. I suggest parsing the path using `urllib.parse` because it will give you an object that will actually parse the url scheme, so you don't need the `S3PREFIX` variable: https://docs.python.org/3/library/urllib.parse.html. Though for windows path, it might not work with `urllib.parse` so you'd need to check for that somehow. 😛

carlosgjs: Good point, added the check via `urlparse`

src/noisepy/seis/utils.py
```diff
@@ -0,0 +1,19 @@
...
+from urllib.parse import urlparse
+
+import fsspec
+
+S3_SCHEME = "s3"
+
+
+def get_filesystem(path: str) -> fsspec.AbstractFileSystem:
+    url = urlparse(path)
+    return fsspec.filesystem(S3_SCHEME, anon=True) if url.scheme == S3_SCHEME else fsspec.filesystem("file")
```
lsetiawan, line 12:
```suggestion
def get_filesystem(path: str, storage_options: dict = {}) -> fsspec.AbstractFileSystem:
    url = urlparse(path)
    if url.scheme == S3_SCHEME:
        storage_options = {'anon': True}
    return fsspec.filesystem(url.scheme, **storage_options)
```

Just a little change to this code. Since this is a "utility" module, I think it should be agnostic in terms of file system, but defaulting to the storage option of `anon=True` for s3. However, maybe you'd like to use this utility again for other file system such as Google Cloud or Azure Blob Storage / Data Lake.

tests/test_utils.py
```diff
@@ -0,0 +1,14 @@
...
+paths = [
+    ("/dir/", "file.txt", "/dir/file.txt"),
+    ("../relative", "file.json", "../relative/file.json"),
+    ("s3://bucket/path", "file.xml", "s3://bucket/path/file.xml"),
+]
+
+
+@pytest.mark.parametrize("path1, path2, expected", paths)
+def test_fs_join(path1: str, path2: str, expected: str):
+    assert expected == fs_join(path1, path2)
```
lsetiawan, line 14: Please add test for `get_filesystem` method.

src/noisepy/seis/scedc_s3store.py
```diff
@@ -27,20 +27,22 @@ class SCEDCS3DataStore(RawDataStore):
...
         Parameters:
-            directory: directory to look for ms files
+            path: path to look for ms files. Can be a local file directory or an s3://... path
             chan_catalog: ChannelCatalog to retrieve inventory information for the channels
             channel_filter: Function to decide whether a channel should be used or not,
                             if None, all channels are used
         """
         super().__init__()
-        self.directory = directory
+        self.fs = get_filesystem(path)
```
lsetiawan, line 39: I was just thinking, is it worth to check for the correct scheme of `s3` if the path is a URL? Seem like that's important if this module is meant to work with `s3` ONLY and not other file systems?

src/noisepy/seis/utils.py
carlosgjs: Thanks, I'll make a small update, add the `anon=True`  only if the the `anon` key is not already there.

src/noisepy/seis/scedc_s3store.py
carlosgjs: No, it's not meant to work exclusively with S3. It only happens to be where we have data now, so I don't think we need to constrain it.

lsetiawan: Okay. In that case, maybe need to rename the module to be `scedc_store` and remove references to just S3?

carlosgjs: Oh, I see what you mean now, yes, the scedc_s3store is specific to S3 and the file naming conventions used there by SCEDC. However, I can see somebody mirroring this data (or a portion) with a different provider (eg. Azure blobs), just like you can download some of it locally. In this case I see more value in allowing different fsspec providers to be used. The class is still meant to follow the schema/naming of the original S3 bucket though, so I think it makes sense for the name to reflect that.

tests/test_utils.py
```diff
@@ -0,0 +1,28 @@
...
+fs_types = [
+    ("s3://bucket/path", type(S3FileSystem())),
+    ("/some/file", type(LocalFileSystem())),
+    ("s3/local/file", type(LocalFileSystem())),
+]
+
+
+@pytest.mark.parametrize("path, fs_type", fs_types)
+def test_get_filesystem(path, fs_type):
+    assert type(get_filesystem(path)) == fs_type
```
lsetiawan, line 28:
```suggestion
fs_types = [
    ("s3://bucket/path", S3FileSystem),
    ("/some/file", LocalFileSystem),
    ("s3/local/file", LocalFileSystem),
]


@pytest.mark.parametrize("path, fs_type", fs_types)
def test_get_filesystem(path, fs_type):
    assert isinstance(get_filesystem(path), fs_type)
```

src/noisepy/seis/scedc_s3store.py
lsetiawan: Gotcha. Okay. In that case this module is appropriately named. I think an explanation somewhere about this naming scheme would be helpful for a user. Thanks for clearing that up 😄

src/noisepy/seis/utils.py
```diff
@@ -0,0 +1,23 @@
...
+
+def get_filesystem(path: str, storage_options: dict = {}) -> fsspec.AbstractFileSystem:
+    url = urlparse(path)
+    # default to anonymous access for S3 if the this is not already specified
+    if url.scheme == S3_SCHEME and ANON_ARG not in storage_options:
+        storage_options = {ANON_ARG: True}
+    return fsspec.filesystem(url.scheme, **storage_options)
+
+
+def fs_join(path1: str, path2: str) -> str:
```
lsetiawan, line 19: Please add some docstring for both functions in this module. Thanks!

carlosgjs:
@lsetiawan take 5!

src/noisepy/seis/channelcatalog.py
```diff
@@ -54,25 +54,27 @@ def get_inventory(self, timespan: DateTimeRange, station: Station) -> obspy.Inve
 
 class XMLStationChannelCatalog(ChannelCatalog):
     """
-    A channel catalog that reads <station>.XML files from a directory
+    A channel catalog that reads <station>.XML files from a directory or an s3://... bucket path.
```
lsetiawan, line 57:
```suggestion
    A channel catalog that reads <station>.XML files from a directory or an s3://... bucket url path.
```

src/noisepy/seis/main.py
```diff
@@ -59,8 +59,10 @@ def get_channel_filter(sta_list: List[str]) -> Callable[[Channel], bool]:
 
 
 def create_raw_store(raw_dir: str, sta_list: List[str], xml_path: str):
+    fs = get_filesystem(raw_dir)
+
     def count(pat):
-        return len(glob(os.path.join(raw_dir, pat)))
+        return len(fs.glob(os.path.join(raw_dir, pat)))
```
lsetiawan, line 65: Does this one matter if it's not using your `fs_join`?

src/noisepy/seis/scedc_s3store.py
```diff
@@ -27,20 +27,22 @@ class SCEDCS3DataStore(RawDataStore):
     # for checking the filename has the form: CIGMR__LHN___2022002.ms
     file_re = re.compile(r".*[0-9]{7}\.ms$", re.IGNORECASE)
 
-    def __init__(self, directory: str, chan_catalog: ChannelCatalog, channel_filter: Callable[[Channel], bool] = None):
+    def __init__(self, path: str, chan_catalog: ChannelCatalog, channel_filter: Callable[[Channel], bool] = None):
         """
         Parameters:
-            directory: directory to look for ms files
+            path: path to look for ms files. Can be a local file directory or an s3://... path
```
lsetiawan, line 33:
```suggestion
            path: path to look for ms files. Can be a local file directory or an s3://... url path
```

lsetiawan (approved):
This now looks good to merge. There are a few minor edits and a small Q. 😄 Thanks for incorporating the changes!

src/noisepy/seis/main.py
carlosgjs: Good catch! it does matter, I'll change it ...

PR #106, opened 2023-05-08
Channel filter helper and S0 client parameter

carlosgjs:
Two small changes

- Allow passing a `client_url_key` parameter to the `download` function, with `"SCEDC"` as the defautl.
- A helper for filtering channels in cross correlation, so a lambda is not needed in the common case. 

I.e., instead of:
```
SCEDCS3DataStore(S3_DATA, catalog, lambda c: c.station.name in stations and c.type.name.startswith("BH"))
```
It can be:
```
SCEDCS3DataStore(S3_DATA, catalog, channel_filter(stations, "BH"))
```


mdenolle (approved):
thank you so much!

PR #111, opened 2023-05-10
add so cal csv file

niyiyu:


niyiyu:
Has been added in #112

PR #112, opened 2023-05-10
update plotting, conj

mdenolle:
basic to remove the most important bugs: np.conj(ff1), plotting issues, and trying out the new mpiexex


niyiyu (approved):
Approved!

mdenolle:
we have to do it because the bug is too big that i absolutely have to fix it

PR #116, opened 2023-05-15
Refactor S1 to use a dataclass for FFT data

carlosgjs:
This is a pre-cursor step to parallelizing the cross-correlation computations. The FFTs/STD and other data about a channel are now put in a data structure (`NoiseFFT`) instead of single arrays. This helps in a few ways:

- The array doesn't have 'holes' if some channel doesn't have data, we just don't store an object for this array. This removes the need for the `fft_flag` array.
- There's less indexing, making the code more readable. E.g. `src_fft.std` vs `fft_std[iiS]`

### DIFF
--- a/src/noisepy/seis/datatypes.py
+++ b/src/noisepy/seis/datatypes.py
@@ -168,5 +168,5 @@ class NoiseFFT:
     fft: np.ndarray
     std: np.ndarray
     fft_time: np.ndarray
-    N: int
-    Nfft2: int
+    SegmentCount: int
+    Length: int
--- a/src/noisepy/seis/datatypes.py
+++ b/src/noisepy/seis/datatypes.py
@@ -157,3 +157,16 @@ class ChannelData:
         self.data = stream[0].data
         self.sampling_rate = stream[0].stats.sampling_rate
         self.start_timestamp = stream[0].stats.starttime.timestamp
+
+
+@dataclass
+class NoiseFFT:
+    """
+    Data class to hold FFT and associated data
+    """
+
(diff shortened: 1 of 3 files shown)


src/noisepy/seis/datatypes.py
```diff
@@ -157,3 +157,16 @@ def __init__(self, stream: obspy.Stream):
         self.data = stream[0].data
         self.sampling_rate = stream[0].stats.sampling_rate
         self.start_timestamp = stream[0].stats.starttime.timestamp
+
+
+@dataclass
+class NoiseFFT:
```
carlosgjs, line 163: @mdenolle - some naming questions:
- Is there a better name for this class? 
- Is `N` the number of segments? If so I'd like to rename it to something line `SegmentCount`. 
- Looks like `Nfft2` is half the spectrum size? If so I'd prefer to return the full size, call it  `SpectrSz` and have the caller half-it as needed.

mdenolle: N is indeed the number of segments or the count of windows. I see in S1 that we also use "nseg_chunk".

The time series length is ``fft_params.cc_len * fft_params.samp_freq``. We fill with zeros before doing the FFT up to ``nnfft`` because it speeds up the FFT.

``NFFT2`` should be ``nnfft//2``. In a way, i agree that we don't need to call. But we should not store the second half of the FFT and only keep in memory the data up to nnfft//2.

mdenolle (approved):
Adding the FFT data class looks good!

aniketfadia96 (approved):
LGTM

PR #117, opened 2023-05-16
Parallelization of S1 using Ray

carlosgjs:
Changes:
- Remove MPI parallelization at the day (time) level and replaced Ray at the fft/pair level
- Disabled compression in the ASDF CC store. I found that the write time is decreased dramatically (20s->1s) while the space savings are very modest (832mb -> 1gb)
- Decreased the compatibility to python `<3.11` since Ray is not yet released for `3.11` (but already in progress).
## Tests on an M2 Max Mac:
Before:
```
2023-05-15 15:51:57,498 INFO utils log TIMING: 2.3873 for get channels
2023-05-15 15:51:59,174 INFO utils log TIMING: 1.6761 for read channel data
2023-05-15 15:52:05,659 INFO utils log TIMING: 6.4848 for preprocess
2023-05-15 15:52:09,590 INFO utils log TIMING: 3.9311 for Compute FFTs
2023-05-15 15:52:47,101 INFO utils log TIMING: 51.9899 for Process the chunk of 2022-01-02T00:00:00+0000 - 2022-01-03T00:00:00+0000
2023-05-15 15:52:47,102 INFO utils log TIMING: 51.9919 for Step 1 in total
```
After:
```
2023-05-16 13:16:50,440 INFO utils log TIMING: 2.3704 for get channels
2023-05-16 13:16:53,092 INFO utils log TIMING: 2.6512 for read channel data
2023-05-16 13:16:54,774 INFO utils log TIMING: 1.6821 for preprocess
2023-05-16 13:16:55,300 INFO utils log TIMING: 0.5260 for Compute FFTs
2023-05-16 13:16:57,604 INFO utils log TIMING: 2.3044 for Created 18 CC tasks
2023-05-16 13:17:00,628 INFO utils log TIMING: 12.5574 for Process the chunk of 2022-01-02T00:00:00+0000 - 2022-01-03T00:00:00+0000
2023-05-16 13:17:00,628 INFO utils log TIMING: 14.5602 for Step 1 in total
```
Note: The increase in the time for `read channel data` is due to the fact that this is the first Ray call, so we pay the initialization cost here. If I repeat this call the time goes down to ~0.5s.

### DIFF
--- a/script/s1perf.sh
+++ b/script/s1perf.sh
@@ -1,4 +1,6 @@
 rm -rf ~/ccfs3tmp
 LOGFILE="log$(date -j +'%Y%m%d_%H%M%S').txt"
 echo Logging to $LOGFILE
-noisepy cross_correlate --raw_data_path ~/s3tmp/scedc/2022/002 --ccf_path ~/ccfs3tmp --freq_norm rma --stations "SBC,RIO,DEV,HEC,RLR,SVD,RPV,BAK" --xml_path ~/s3tmp/FDSNstationXML/ 2>&1 | tee $LOGFILE
+# STATIONS="SBC,RIO,DEV,HEC,RLR,SVD,RPV,BAK,CBC,CCA,CCC,CDD,HLL,HOL,AGM,AGO"
+STATIONS="SBC,RIO,DEV,HEC,RLR,SVD,RPV,BAK"
+noisepy cross_correlate --raw_data_path ~/s3tmp/scedc/2022/002 --ccf_path ~/ccfs3tmp --freq_norm rma --stations "$STATIONS" --xml_path ~/s3tmp/FDSNstationXML/ 2>&1 | tee $LOGFILE
--- a/script/s1perf.sh
+++ b/script/s1perf.sh
@@ -1,5 +1,6 @@
 rm -rf ~/ccfs3tmp
-LOGFILE="log$(date -j +'%Y%m%d_%H%M%S').txt"
+mkdir -p $HOME/logs
+LOGFILE="$HOME/logs/log$(date -j +'%Y%m%d_%H%M%S').txt"
 echo Logging to $LOGFILE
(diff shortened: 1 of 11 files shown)


src/noisepy/seis/datatypes.py
```diff
@@ -168,5 +168,5 @@ class NoiseFFT:
     fft: np.ndarray
     std: np.ndarray
     fft_time: np.ndarray
-    SegmentCount: int
-    Length: int
+    segment_count: int
```
mdenolle, line 171: could we call that "window_count" i think seismologists prefer the use of "window" rather than "segment"

script/s1perf.sh
```diff
@@ -0,0 +1,7 @@
+rm -rf ~/ccfs3tmp
+mkdir -p $HOME/logs
+LOGFILE="$HOME/logs/log$(date -j +'%Y%m%d_%H%M%S').txt"
+echo Logging to $LOGFILE
+# STATIONS="SBC,RIO,DEV,HEC,RLR,SVD,RPV,BAK,CBC,CCA,CCC,CDD,HLL,HOL,AGM,AGO"
+STATIONS="SBC,RIO,DEV,HEC,RLR,SVD,RPV,BAK"
+noisepy cross_correlate --raw_data_path ~/s3tmp/scedc/2022/002 --ccf_path ~/ccfs3tmp --freq_norm rma --stations "$STATIONS" --xml_path ~/s3tmp/FDSNstationXML/ 2>&1 | tee $LOGFILE
```
mdenolle, line 7: I do not see any problem with te lines in S1_fft_cc_MPI. I wonder if the result of this test gives data that looks like cross correlations. When we use the plotting commands in the get_started tutorial:
file = os.path.join(cc_data_path, '2019_02_01_00_00_00T2019_02_01_12_00_00.h5')
plotting_modules.plot_substack_cc(file,0.1,1,200,False)
does it plot what it did before?

carlosgjs: I ran the get_started tutorial and as far as I can tell the plots look the same as before. Couple of examples:

image
image

Can you confirm?

PR #118, opened 2023-05-17
Use ConfigParameters in S2

carlosgjs:


mdenolle (approved):
thanks for updating S2

PR #119, opened 2023-05-17
Remove old/dead code

carlosgjs:
I went over the files under `scripts`, which are currently not part of the noisepy package. These are the ones that appear to be either one-off test scripts or earlier/duplicated versions of the code.

Let me know if there's long term value in keeping some of these, or if there are additional files we can remove. Having less code to maintain will allow us to move faster.


mdenolle (approved):
thank you for removing the unused tests

PR #120, opened 2023-05-18
Use CC store in S2

carlosgjs:
Changes

- Similar to the change made in S1, S2 no uses the CrossCorrelationDataStore instead of directly accessing the ASDF files.
- Refactored ASDF stores to share functionality between the raw and CC stores
- Add unit tests for the new reading functions in the CC store
- Add a constructor for `Station` to use when lat/lon/elev are not known. Previously `0` and `-1` were inconsistently used as missing values, neither of which was good.
- Fix problem with double initialization of ray.

TODO: Implement a similar store for writing the stacked data

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -7,7 +7,7 @@
       "request": "launch",
       "module": "pytest",
       "args": [
-        "${workspaceFolder}/tests/test_scedc_s3store.py"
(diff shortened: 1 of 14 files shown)


mdenolle (approved):
extensive refactoring, thank you so much! I reviewed and am excited to test the new notebook

PR #126, opened 2023-05-25
Implement a Zarr CrossCorrelationDataStore 

nrich20:
Implemented a Zarr storage module for saving cross correlation data in the Zarr file format.


src/noisepy/seis/zarrstore.py
```diff
@@ -0,0 +1,154 @@
...
+            logger.info(f"Cross-correlation {station_pair} and {channel_pair} already exists")
+        return contains
+
+    def save_parameters(self, parameters: ConfigParameters):
+        fc_metadata = os.path.join(self.directory, "fft_cc_data.json")
+
+        fout = open(fc_metadata, "w")
+        # WIP actually serialize this
+        fout.write(str(parameters.__dict__))
+        fout.close()
```
carlosgjs, line 91: No longer needed 
```suggestion
```

src/noisepy/seis/zarrstore.py
```diff
@@ -0,0 +1,154 @@
...
+        with zarr.open(store, mode="a") as ccf_ds:
+            ccf_ds.attrs.setdefault("auxiliary_data", {}).setdefault(data_type, {})
+            ccf_ds.attrs["auxiliary_data"][data_type][path] = {"parameters": params}
+            ccf_ds.create_dataset(f"auxiliary_data/{data_type}/{path}", data=data)
+
+    def _get_station_pair(self, src_chan: Channel, rec_chan: Channel) -> str:
+        return f"{src_chan.station}_{rec_chan.station}"
+
+    def _get_channel_pair(self, src_chan: Channel, rec_chan: Channel) -> str:
+        return f"{src_chan.type.name}_{rec_chan.type.name}"
```
carlosgjs, line 148: Can you move these into `stores.py` and share them between `ASDFCCStore` and `ZarrCCStore` to avoid duplication? Note that these now take `Station` and `ChannelType` args respectively.

src/noisepy/seis/zarrstore.py
```diff
@@ -0,0 +1,154 @@
...
+    def mark_done(self, timespan: DateTimeRange):
+        self._add_aux_data(timespan, {}, PROGRESS_DATATYPE, DONE_PATH, np.zeros(0))
+
+    def is_done(self, timespan: DateTimeRange):
+        done = self._contains(timespan, PROGRESS_DATATYPE, DONE_PATH)
+        if done:
+            logger.info(f"Timespan {timespan} already computed")
+        return done
+
+    def read(self, chan1: Channel, chan2: Channel, start: datetime, end: datetime) -> np.ndarray:
```
carlosgjs, line 119: take a look at the latest definition in `stores.py`, there's also an implementation in `asdfstore.py`

src/noisepy/seis/zarrstore.py
```diff
@@ -0,0 +1,154 @@
...
+
+from . import noise_module
+from .constants import DATE_FORMAT, DONE_PATH, PROGRESS_DATATYPE
+from .datatypes import Channel, ChannelData, ChannelType, ConfigParameters, Station
+from .stores import CrossCorrelationDataStore, RawDataStore
+
+logger = logging.getLogger(__name__)
+
+
+class ZarrRawDataStore(RawDataStore):
```
carlosgjs, line 21: We don't have a requirement for a ZarrRawDataStore at this time.

src/noisepy/seis/zarrstore.py
```diff
@@ -0,0 +1,154 @@
...
+        try:
+            inventory = arr[name]["StationXML"]
+            sta, net, lon, lat, elv, loc = noise_module.sta_info_from_inv(inventory)
+            return Station(net, sta, lat, lon, elv, loc)
+        except Exception as e:
+            logger.warning(f"Missing StationXML for station {name}. {e}")
+            return None
+
+
+class ZarrCCStore(CrossCorrelationDataStore):
```
carlosgjs, line 67: Please add unit testing for this class. Take a look at `test_asdfstore.py:test_ccstore`. I think we can reuse the same test code for both the ASDF and Zarr CC stores. I.e. you can make a helper test method that takes an abstract `CrossCorrelationStore` as an argument and then write two tests that call it with different stores. E.g.
```

def _ccstore_test_helper(ccstore: CrossCorrelationStore):
... current test code

def test_asdfccstore(asdfstore: ASDFCCStore):
    _ccstore_test_helper(asdfstore)

def test_zarrccstore(zarrstore: ZarrCCStore):
    _ccstore_test_helper(zarrstore)
```
You would need to add a new test fixture like `ccstore` but for Zarr.

carlosgjs (changes requested):

PR #129, opened 2023-05-26
Update and test notebooks

carlosgjs:
- Automatically run both notebooks as part of the PR checks. Otherwise it's easy to forget to test the notebooks manually. The output is saved with the results ([example](https://github.com/mdenolle/NoisePy/suites/13181985064/artifacts/717447013))
- Update the notebooks to use a relative directory for data. This helps by not needing to change the path when running locally vs in colab. The target directories have been added to the `.gitignore` to avoid accidentally committing these files
- Couple of small updates to the notebooks


lsetiawan (commented):
This is really great @carlosgjs! You really put in so much work in trying to get a nice way to put testing for your tutorial notebooks. I see that there are some manual setup for each notebook to run that you'd have to configure within the gh action yaml. Some ideas that I have to add the functionality to this are:

1. Merge tutorials with docs
2. Use [jupyterbook](https://jupyterbook.org/en/stable/intro.html) to build both these notebooks and docs. Note: Jupyterbook does an automatic run of the notebooks to ensure that they work and will build a directory of htmls automatically. This in turn can be hosted someplace and you'd even be able to have a nice html previews of the tutorials in the future!
3. Add a notebook pre-commit hook for all the typical code checks with: https://github.com/nbQA-dev/nbQA

Let me know what you think of those ideas? I can definitely help you set it up. 😄

mdenolle (approved):

PR #131, opened 2023-06-01
Use Jupyterbook to test/publish tutorial notebooks

carlosgjs:
A workflow that builds the HTML pages. On pull-requests, it uploads the pages as artifacts and on the PR. On pushes to master it will publish them as GitHub pages.


lsetiawan (approved):
This looks good to me in terms of the setup and everything. One thing though, the explicit pip install should probably be removed to not clutter the rendered `ipynb` and overwrite any installation. For example, I had `grpcio` installed via pip from that line in the notebook, which I don't want, since the `ray` package actually wants the `grpcio` from conda. So having that line is problematic.

PR #138, opened 2023-06-09
Support custom file names in XMLStationChannelCatalog

carlosgjs:
- Allow for passing a custom format string to build a file name from a `Station`
- In S1, abort if the stations do not have lat/lon/elevation information
- Added unit tests


mdenolle (approved):
looks good!

PR #139, opened 2023-06-12
RFC: PNWStore - Query DB during read_data()

carlosgjs:
While debugging with Kuan-Fu we ran into an issue because of the `byteoffset`,`bytes` and `filename` properties that were being dynamically added to the Channels. The issue is that once we use the `XMLStationChannelCatalog.get_full_channels()` then the channels returned from `get_channels()` are new objects that don't have the dynamic properties (e.g. `byteoffset`).

The proposal here is to read these properties from the DB at they time they are needed. This preserves the use of the `Channel` objects to just their strongly typed properties. If this turns out to be slow we can cache these properties in the store itself.

Note that this code is untested (since I don't have access to the DB) and would still need to be properly factored.


src/noisepy/seis/pnwstore.py
```diff
@@ -77,20 +78,23 @@ def __init__(
...
 
     def _load_channels(self, full_path: str, ch_filter: Callable[[Channel], bool]):
-        _, _, _, _, _, net, year, doy, _ = full_path.split('/')
+        _, _, _, _, _, net, year, doy, _ = full_path.split("/")
         sqlite = f"/fd1/yiyu_data/PNWstore_sqlite/{year}.sqlite"
         db = sqlite3.connect(sqlite)
         cursor = db.cursor()
-        rst = cursor.execute(f"SELECT network, station, channel, location, filename, byteoffset, bytes " \
-                              f"FROM tsindex WHERE filename LIKE '%%/{net}/{year}/{doy}/%%'")
+        rst = cursor.execute(
```
carlosgjs, line 85: Here we query just the information to build the channel list

src/noisepy/seis/pnwstore.py
```diff
@@ -124,20 +128,34 @@ def get_timespans(self) -> List[DateTimeRange]:
...
-        filename = fs_join(
-            self.paths[timespan.start_datetime], f"{chan_str}"
+        assert timespan.start_datetime.year == timespan.end_datetime.year, "Did not expect timespans to cross years"
+        year = timespan.start_datetime.year
+        doy = timespan.start_datetime.timetuple().tm_yday
+
+        # TODO: Refactor and deduplicate the DB querying code with _load_channels()
+        sqlite = f"/fd1/yiyu_data/PNWstore_sqlite/{year}.sqlite"
+        db = sqlite3.connect(sqlite)
+        cursor = db.cursor()
```
carlosgjs, line 138: We query for byteoffset and bytes during the read

src/noisepy/seis/pnwstore.py
```diff
@@ -159,12 +177,12 @@ def _parse_channel(query: tuple) -> Channel:
...
         location = query[3]
-        c =  Channel(
+        c = Channel(
             ChannelType(channel, location),
             # lat/lon/elev will be populated later
             Station(network, station, location=location),
         )
-        c.filename = query[4]
-        c.byteoffset = query[5]
-        c.bytes = query[6]
```
carlosgjs, line 169: these are not properties of a Channel object.

kuanfufeng (approved):

PR #147, opened 2023-07-06
Make MPI an optional dependency

carlosgjs:
Having MPI as a core dependency presented several difficulties. It requires an MPI implementation, which is often platform dependent and takes longer to install. Installing MPI was one of the difficulties people faced on the last workshop. Furthermore, it's typically not required in cloud or single machine deployments.

This PR makes it an optional dependency by:
- Adding a `Scheduler`  abstraction to factor out the MPI/non-MPI details from the main computation
- Making mpi an optional dependency (ie `[mpi]`)
- Adding support for the CLI through a `--mpi` argument
- Updating the CI testing to not require MPI for the bulk of the tests (faster) but adding a specific MPI test

### DIFF
--- a/src/noisepy/seis/S0A_download_ASDF_MPI.py
+++ b/src/noisepy/seis/S0A_download_ASDF_MPI.py
@@ -19,6 +19,7 @@ from noisepy.seis.datatypes import ConfigParameters
 from noisepy.seis.utils import TimeLogger
 from . import noise_module
 from obspy.clients.fdsn import Client
+from obspy.clients.fdsn.header import FDSNNoDataException
 
 logger = logging.getLogger(__name__)
 if not sys.warnoptions:
@@ -281,6 +282,7 @@ def download_stream(
     client = Client(prepro_para.client_url_key, timeout=15)
     retries = 5
     while retries > 0:
+        retries -= 1
         try:
             tr = client.get_waveforms(
                 network=net,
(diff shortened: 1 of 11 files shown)


src/noisepy/seis/S0A_download_ASDF_MPI.py
```diff
@@ -292,6 +292,7 @@ def download_stream(
             )
         except Exception as e:
             logger.warning(f"{e} for get_waveforms({sta}.{chan})")
+            continue
```
carlosgjs, line 295: Unrelated but fixing this here opportunistically

src/noisepy/seis/scheduler.py
```diff
@@ -0,0 +1,91 @@
...
+            for i in range(shared_vars):
+                # receive shared variables
+                vars.append(self.comm.bcast(None, root=self.root))
+            logger.debug(f"RANK {rank}, vars = {vars}")
+            return vars
+
+    def get_indices(self, items: list) -> List[int]:
+        rank = self.comm.Get_rank()
+        size = self.comm.Get_size()
+        rank = self.comm.Get_rank()
```
aniketfadia96, line 84:
```suggestion
        
```

aniketfadia96: `rank` is initialized on line 82.

aniketfadia96 (changes requested):
Added one minor comment. LGTM overall.

src/noisepy/seis/scheduler.py
carlosgjs: Good catch!

aniketfadia96 (approved):

PR #148, opened 2023-07-06
Support using MPI for CC

carlosgjs:
This PR builds on #147 to add back support for MPI in the cross-correlation step.


aniketfadia96 (approved):
LGTM

PR #153, opened 2023-07-19
Update plot_substack_cc to use CC store

carlosgjs:
We need to update the plotting functions to use the store abstraction instead of reading ASDF files directly. This will enable using other formats, such as Zarr.

Changes:
- Support read operations on the StackStore
- Add unit tests for StackStore
- Update plotting module functions to use stores
- Update notebooks


aniketfadia96 (approved):
LGTM.

PR #158, opened 2023-07-24
Support Zarr format from the CLI

carlosgjs:
- Adds a `--format zarr|asdf` argument to the CLI
- Additional timing measurement and logging
- Update the test matrix to:
    - Include for format (zarr vs asdf)
    - Be more sparse. All types of normalization and stacking are tested, but not all combinations, which is overkill

### DIFF
--- a/.github/workflows/precommit.yaml
+++ b/.github/workflows/precommit.yaml
@@ -16,6 +16,6 @@ jobs:
       with:
         python-version: 3.8
     - name: Install pre-commit
-      run: pip install pre-commit==3.3
+      run: pip install pre-commit==3.3.3
     - name: Run pre-commit checks
       run: pre-commit run --all-files --show-diff-on-failure
--- a/.github/workflows/precommit.yaml
+++ b/.github/workflows/precommit.yaml
@@ -16,6 +16,6 @@ jobs:
       with:
         python-version: 3.8
     - name: Install pre-commit
-      run: pip install pre-commit
+      run: pip install pre-commit==3.3
     - name: Run pre-commit checks
       run: pre-commit run --all-files --show-diff-on-failure
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -42,18 +42,18 @@ jobs:
       fail-fast: true
       matrix:
         python_version: ['3.8', '3.10']
-        method: [linear]
+        method: [linear,robust, nroot, selective, auto_covariance, pws]
         freq_norm: [rma]
         format: [asdf]
         include:
-        - method: [robust, nroot, selective, auto_covariance, pws]
-          python_version: ['3.8', '3.10']
-          freq_norm: [no]
-          format: [zarr]
-        - method: [robust]
-          python_version: ['3.8', '3.10']
-          freq_norm: [phase_only]
-          format: [zarr]
+        - method: linear
+          python_version: '3.10'
+          freq_norm: no
+          format: zarr
+        - method: robust
+          python_version: '3.8'
+          freq_norm: phase_only
+          format: zarr
     runs-on: ubuntu-22.04
     needs: s0_download
     steps:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -51,7 +51,7 @@ jobs:
           freq_norm: no
           format: zarr
         - method: robust
-          python_version: '3.8'
+          python_version: '3.9'
           freq_norm: phase_only
           format: zarr
     runs-on: ubuntu-22.04
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -75,7 +75,7 @@ jobs:
     - name: Test Cross-Correlation (S1)
       working-directory: ./src
       run: |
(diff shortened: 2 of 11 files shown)


.github/workflows/test.yaml
```diff
@@ -42,8 +42,18 @@ jobs:
...
+        method: [linear]
+        freq_norm: [rma]
+        format: [asdf]
+        include:
+        - method: [robust, nroot, selective, auto_covariance, pws]
+          python_version: ['3.8', '3.10']
+          freq_norm: [no]
+          format: [zarr]
+        - method: [robust]
+          python_version: ['3.8', '3.10']
```
lsetiawan, line 54: Any reason why Python 3.9 is not included in the testing?

.github/workflows/precommit.yaml
```diff
@@ -16,6 +16,6 @@ jobs:
       with:
         python-version: 3.8
     - name: Install pre-commit
-      run: pip install pre-commit
+      run: pip install pre-commit==3.3
```
lsetiawan, line 19: Maybe make this pinning same as in `pyproject.toml`?

src/noisepy/seis/main.py
```diff
@@ -37,6 +38,11 @@ class Command(Enum):
     ALL = 4
 
 
+class DataFormat(Enum):
+    ZARR = 1
+    ASDF = 2
```
lsetiawan, line 43: Are you actually using the values of this enum anywhere? Would it be better to be the lower name values instead? Seems like that's the only thing you're using it for, might save some of the `.lower` computation.

.github/workflows/test.yaml
carlosgjs: Just to keep the matrix smaller since I was doing the full matrix testing. I opted for testing the min and max of supported versions. But actually now that I discovered the `include` argument I can add `3.9` for a single configuration as a good sanity check.

src/noisepy/seis/main.py
carlosgjs: It's a habit since an int comparison is generally much faster than a string comparison and has fewer gotchas (like casing and culture). But I agree in this case the string value makes things simpler and won't make a difference perf-wise. I'll update ...

.github/workflows/precommit.yaml
carlosgjs: yep, already done

.github/workflows/test.yaml
```diff
@@ -42,8 +42,18 @@ jobs:
       fail-fast: true
       matrix:
         python_version: ['3.8', '3.10']
```
lsetiawan, line 44: Does this software not work with python 3.9? I'm curious why that's not included.

tests/test_cli.sh
```diff
@@ -1,14 +1,42 @@
...
+       echo "Missing FORMAT argument (zarr or asdf)"
+       exit 1
+fi
+echo "FORMAT is $FORMAT"
+LOGFILE="$HOME/logs/log_${FORMAT}_$(date -j +'%Y%m%d_%H%M%S').txt"
+STATIONS=ARV,BAK
+START=2019-02-01T00:00:00
+END=2019-02-01T01:00:00
+INC_HOURS=1
+
```
carlosgjs, line 16: Results for this "bigger test":
image
This is running on my Mac, so local SSD I/O. All times in seconds.
TL; DR: Zarr is slower for writing, but faster for reading. But also note that there is not a huge difference in the overall time. However, this may change with: a) parallelization of stacking and b) cloud storage

.github/workflows/test.yaml
carlosgjs: same as above

src/noisepy/seis/S2_stacking.py
```diff
@@ -250,6 +253,7 @@ def write_stacks(comp: str, tparameters: Dict[str, Any], stacks: List[Tuple[Stac
                     ]
                     write_stacks(comp, tparameters, stacks)
         tlog.log(f"stack/rotate all station pairs {pairs_all[ipair]}", t_load)
+        tlog.log_raw("store.append", t_append)
```
carlosgjs, line 256: @niyiyu

src/noisepy/seis/S1_fft_cc_MPI.py
```diff
@@ -142,11 +144,15 @@ def init() -> List:
...
 
+        t_append = 0
         for t in as_completed(tasks):
             # Use as_completed so we can start saving results to the store
             # while other computations are still running
             src_chan, rec_chan, parameters, corr = t.result()
+            t_start = time.time()
             cc_store.append(ts, src_chan, rec_chan, parameters, corr)
+            t_append += time.time() - t_start
+        tlog.log_raw("Append to store", t_append)
```
carlosgjs, line 155: @niyiyu

.github/workflows/test.yaml
carlosgjs: Added one test with 3.9

niyiyu (approved):

.pre-commit-config.yaml
```diff
@@ -12,6 +12,7 @@ repos:
       - id: check-json
       - id: check-yaml
       - id: pretty-format-json
+        files: \.(json|json5|js)$
```
lsetiawan, line 15: Maybe instead of specifying what to include, exclude the `.ipynb` file format since we don't want that to be prettified anyways.

PR #159, opened 2023-07-25
Parallelize stacking

carlosgjs:
Tested with the following parameters:
```
STATIONS=ADO,ALP,ARV,AVM,BAK,BAR,BBR,BBS,BC3,BCW
START=2019-02-01T00:00:00
END=2019-02-05T00:00:00
INC_HOURS=24
```
Total stacking time went down from `173.19s` to `26.86`.

### DIFF
--- a/src/noisepy/seis/S2_stacking.py
+++ b/src/noisepy/seis/S2_stacking.py
@@ -1,16 +1,17 @@
 import logging
 import sys
 import time
+from concurrent.futures import ProcessPoolExecutor
 from typing import Any, Dict, List, Tuple
 
 import numpy as np
 import pandas as pd
 from datetimerange import DateTimeRange
 
-from noisepy.seis.datatypes import ConfigParameters, StackMethod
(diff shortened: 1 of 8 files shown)


src/noisepy/seis/S2_stacking.py
```diff
@@ -49,8 +50,49 @@ def stack(
...
+
+        if len(timespans) == 0 or len(pairs_all) == 0:
+            raise IOError("Abort! no available CCF data for stacking")
+        return timespans, pairs_all
+
+    timespans, pairs_all = scheduler.initialize(initializer, 2)
+
+    # Get the pairs that need to be processed by this node
+    pairs_node = [pairs_all[i] for i in scheduler.get_indices(pairs_all)]
+
```
carlosgjs, line 70: The main change is here, using the executor to run all the jobs in parallel.

The bulk of the changes below are: a) a change in indentation since most lines went from being inside a loop to being just part of a function, and b) `continue` -> `return`

niyiyu (approved):

PR #160, opened 2023-07-27
Support for S3 and Zarr

carlosgjs:
Add `storage_options` to ConfigParameters to be able to specify authentication for S3 and other protocols. E.g.:
```
storage_options:
  s3:
    anon: false
    profile: "my_aws_test_profile"
```


aniketfadia96 (approved):
LGTM

Issue #18, opened 2020-06-09
mismatch with documentation for `Testing`


aimanyongki:
The documentations for `Testing` either on [https://noise-python.readthedocs.io/en/latest/installation.html#testing](https://noise-python.readthedocs.io/en/latest/installation.html#testing) or [https://github.com/mdenolle/NoisePy#short-tutorial](https://github.com/mdenolle/NoisePy#short-tutorial) suggest running  
`$ python S0_download_ASDF_MPI.py` 
`$ python S1_fft_cc.py`
`$ python S2_stacking.py`  
to make sure the scripts all pass successfully. 

However, those scripts under `NoisePy/src/` got new names for:
`S0A_dowload_ASDF_MPI.py`
`S1_fft_cc_MPI.py`

Also, running the `S0A_dowload_ASDF_MPI.py` results in an error with the current `rootpath`:
`OSError: file /Users/chengxin/Documents/NoisePy_example/AZ/RAW_DATA/station.txt not exist! double check!`

chengxinjiang:
Hi @aimanyongki,

Many thanks for your interests in using NoisePy as well as the comments on the name discrepancies. We are currently in a phase of one major revision on the codes/documentation, and is aware of this problem. So they should be taken into account into consideration. Thank you for pointing this problem out here.

To run the script of NoisePy, the users are expected to change the parameters such as rootpath in order to find their own targeted path. Let me know how it works after the change of rootpath parameter.

chengxinjiang:
This issue has been fixed now. Thanks again for your comment.

marker07:
$ python3 S1_fft_cc_MPI.py 

==> and then i get 

File "S1_fft_cc_MPI.py", line 354
    if (station[iiR]~=station[iiS]):continue
                    ^
SyntaxError: invalid syntax

==> so i change the modelu tobe :
     if (station[iiR]==station[iiS]):continue

==> but i get some error again 

found 41 stations in total
working on station ARV and trace bhe_00
N and Nfft are 60 (proposed 60),36000 (proposed 18000)
Traceback (most recent call last):
  File "S1_fft_cc_MPI.py", line 298, in <module>
    fft_array[iii] = data.reshape(data.size)
ValueError: could not broadcast input array from shape (1080000,) into shape (540000,)

PR #162, opened 2023-08-01
Noise module updates

Paspachu:
Updated noise_module to use strongly typed parameters class, which is the Issue #18 in CodeUW (https://github.com/uw-ssec/codeuw/issues/18#issue-1694818868).

All parameter loading parts are deleted and variable access is replaced with direct field access.

Might want to change the comments about fft_para since now the variable is a class not a dictionary.

### DIFF
--- a/src/noisepy/seis/noise_module.py
+++ b/src/noisepy/seis/noise_module.py
@@ -533,7 +533,7 @@ def cut_trace_make_stat(fc_para: ConfigParameters, ch_data: ChannelData):
     return trace_stdS, dataS_t, dataS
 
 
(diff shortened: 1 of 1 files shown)


Paspachu:
In the test file, test_whiten.py, the variable fft_para was defined as a dictionary, so I changed it to declare as a ConfigParameters class. Note that instead of fft_para.dt, fft_para.samp_freq is initialized since the dt is a method inside the class which depends on samp_freq value.

I also changed the comment to note that fft_para is ConfigParameters class not dictionary anymore.

Thank you!

carlosgjs (approved):
Looks great!

PR #164, opened 2023-08-01
Add single_freq parameter, bug fixes

carlosgjs:
In some datasets (e.g. SCEDC S3 bucket) the data comes in multiple sampling rates and we need a way of selecting one. So the desired sampling rate is passed as a config parameter and we find the  sampling rate available in the data that is the closest (but equal or greater) than the config. Then we use only the data at that exact frequency.

However, in other scenarios we want to use all the data, as long as its above the desired sampling rate. This PR adds a configuration parameter to control this.

It also adds a couple of checks to avoid crashing when there's missing data.

### DIFF
--- a/src/noisepy/seis/S1_fft_cc_MPI.py
+++ b/src/noisepy/seis/S1_fft_cc_MPI.py
@@ -94,7 +94,9 @@ def cross_correlate(
         )
 
         tlog.log("get channels")
-        ch_data_tuples = _read_channels(executor, ts, raw_store, all_channels, fft_params.samp_freq)
+        ch_data_tuples = _read_channels(
+            executor, ts, raw_store, all_channels, fft_params.samp_freq, fft_params.single_freq
+        )
         # only the channels we are using
 
         if len(ch_data_tuples) == 0:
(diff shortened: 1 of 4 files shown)


src/noisepy/seis/S1_fft_cc_MPI.py
```diff
@@ -300,31 +306,39 @@ def compute_fft(fft_params: ConfigParameters, ch_data: ChannelData) -> NoiseFFT:
...
+    store: RawDataStore,
+    channels: List[Channel],
+    samp_freq: int,
+    single_freq: bool = True,
 ) -> List[Tuple[Channel, ChannelData]]:
     ch_data_refs = [executor.submit(store.read_data, ts, ch) for ch in channels]
     ch_data = _get_results(ch_data_refs)
     tuples = list(zip(channels, ch_data))
-    return _filter_channel_data(tuples, samp_freq)
+    return _filter_channel_data(tuples, samp_freq, single_freq)
```
niyiyu, line 319: Thanks Carlos,

If I understand this this function right, it returns the channels at the frequency closest to `samp_freq` when `single_freq = True`. While `single_freq = False`, it returns all channels with frequency greater than `samp_freq`. By default, the `fft_param` set this switch as True.

Is there any test needs to be added accordingly?

niyiyu: ha I see the tests below now. Looks great!

niyiyu (approved):

PR #167, opened 2023-08-04
Fix PNWDataStore too many files open error

niyiyu:
This PR fix the problem that when a day-long mSEED contains too many gaps, the file system could not handle them well.

### DIFF
--- a/src/noisepy/seis/pnwstore.py
+++ b/src/noisepy/seis/pnwstore.py
@@ -108,6 +108,7 @@ class PNWDataStore(RawDataStore):
             logger.warning(f"Could not find file {timespan}/{chan} in the database")
             return ChannelData.empty()
         elif len(rst) > 10:
+            # skip if stream has more than 10 gaps
             logger.warning(f"Too many gaps (>10) from {timespan}/{chan}")
             return ChannelData.empty()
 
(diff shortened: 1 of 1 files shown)


niyiyu:
Something haven't been implemented are that: 
- When merging the streams, should we fill np.nan, or 0
- Should we skip those mSEED with too many gaps? (gaps more than 100?)
Maybe @mdenolle have thoughts?

mdenolle:
Oh great! Glad that it's not a system issue.
1. filling with zeros might be best, provided that we apply tapers to all segments of the streams.
2. past 10 gaps, i would skip.

niyiyu:
Thanks @mdenolle

I add a 3% taper to each of the segment before merging them. For stream with more than 10 gaps, I simply skip them. Maybe these two parameters could be further moved into the default config. Other than that, this PR is really to go.

Cheers,
Yiyu

mdenolle (approved):
looks good!

src/noisepy/seis/pnwstore.py
```diff
@@ -101,12 +101,16 @@ def read_data(self, timespan: DateTimeRange, chan: Channel) -> ChannelData:
             f"FROM tsindex WHERE network='{chan.station.network}' AND station='{chan.station.name}' "
             f"AND channel='{chan.type.name}' and location='{chan.station.location}' "
             f"AND filename LIKE '%%/{chan.station.network}/{year}/{doy}/%%'"
+            "ORDER BY byteoffset ASC"
```
carlosgjs, line 104: @niyiyu does this need a space at the end of the previous line? `.../%%' "`

niyiyu: Oh seems I missed the space here. However, they seem giving the same results : `.../%%'ORDER BY`..😂
Do you think it worth another PR to add this space?

PR #183, opened 2023-08-18
Improvements to docker publishing

carlosgjs:
- Simplified docker build by using python base image. Since MPI is now optional and not used in cloud deployments the build is much simpler
- Added scripts and docker file for building and publishing from a local repo during development (for fast iteration on AWS)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,4 @@
-FROM --platform=$TARGETPLATFORM ubuntu:22.04
-
-# Avoid timezone prompts during python installation
-ENV DEBIAN_FRONTEND=noninteractive
-
-RUN apt update && \
-    apt install -y libopenmpi-dev && \
-    apt install -y python3.10 && \
-    apt install -y python3-pip && \
-    pip3 install --upgrade pip
+FROM --platform=$TARGETPLATFORM python:3.10.12
 
 ARG VERSION
 RUN pip3 install noisepy-seis==${VERSION}
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
-FROM --platform=$TARGETPLATFORM python:3.10.12
+ARG PYTHON_VERSION=3.10.12
+FROM --platform=$TARGETPLATFORM python:${PYTHON_VERSION}
 
 ARG VERSION
 RUN pip3 install noisepy-seis==${VERSION}
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,16 @@
+FROM --platform=$TARGETPLATFORM python:3.10.12
+
+# This is a version of Dockerfile to build from a local wheel
+# Build with:
+# script/docker_dev.sh
+
+# Split out the install of dependencies and the noisepy code so
+# the dependencies layer gets cached. Much faster build times this way.
+COPY ./dist/_reqs.txt .
+RUN pip install -r _reqs.txt
+
+COPY ./dist/*.whl .
+RUN pip install *.whl
+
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -1,13 +1,4 @@
...
-
-# Avoid timezone prompts during python installation
-ENV DEBIAN_FRONTEND=noninteractive
-
-RUN apt update && \
-    apt install -y libopenmpi-dev && \
-    apt install -y python3.10 && \
-    apt install -y python3-pip && \
-    pip3 install --upgrade pip
+FROM --platform=$TARGETPLATFORM python:3.10.12
```
lsetiawan, line 1: Maybe make the python version tag an argument here as well so you can specify the python version from the outside easily.

Dockerfile.dev
```diff
@@ -0,0 +1,16 @@
+FROM --platform=$TARGETPLATFORM python:3.10.12
```
lsetiawan, line 1: Same as above, make python version an arg

Dockerfile.dev
```diff
@@ -0,0 +1,16 @@
...
+
+# This is a version of Dockerfile to build from a local wheel
+# Build with:
+# script/docker_dev.sh
+
+# Split out the install of dependencies and the noisepy code so
+# the dependencies layer gets cached. Much faster build times this way.
```
lsetiawan, line 8: Clever! 😜

Dockerfile.dev
```diff
@@ -0,0 +1,16 @@
...
+COPY ./dist/_reqs.txt .
+RUN pip install -r _reqs.txt
+
+COPY ./dist/*.whl .
+RUN pip install *.whl
```
lsetiawan, line 13: If you've already installed dependencies, you can just ignore here.
```suggestion
RUN pip install --no-deps *.whl
```

script/docker_dev.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# Script to publish a docker container from the current state of the repo.
+# Useful for iterating on AWS
+rm dist/*.whl
+hatch build -t wheel
+python script/gen_req.py dist/*.whl > ./dist/_reqs.txt
```
lsetiawan, line 6: Okay, so I found that you can just use [`pip-tools`](https://pip-tools.readthedocs.io) rather than your own package to ensure that you get the full dependencies by reading in from `pyproject.toml`, leaving your wheel file alone.

Seems like there's a whole discussions (https://github.com/pypa/pip/issues/11584, ) on being able to pip install only dependencies, especially from a `pyproject.toml`. Someone even have the same exact use case as what you're doing here (https://github.com/pypa/pip/issues/11440).

```suggestion
pip install pip-tools # install pip tools
pip-compile --output-file=./dist/_reqs.txt pyproject.toml # export deps to reqs file
```

carlosgjs: Good to know. Unfortunately, its does the full dependency resolution, so it takes ~10 secs vs 4ms for the python script.

lsetiawan: Is the 4ms includes the docker build step as well? If that's the case, then your current way is probably good. I figured if you get the resolution ahead of time, then the docker build will be a lot faster since pip wouldn't have to resolve then in https://github.com/mdenolle/NoisePy/pull/183/files#diff-86930c95a19b82f7e64a962a0053d44e855824813019b3698eae4917a90cdcacR11

lsetiawan (approved):
LGTM! That addition for `PYTHON_VERSION` looks great! Thanks

PR #184, opened 2023-08-22
S1 fixes for larger workloads on AWS

### LINKED ISSUE #173: Support adding stations to an existing run
Use case

1. Compute CC for some stations:
```
noisepy cross_corrlate stations="ARV,BAC" --start=2022-02-02 --end=2022-02-03 --ccf_path="~/ccf"
```
2. Re-run the command, but add a station (`ALP`)
```
noisepy cross_corrlate stations="ARV,BAC,ALP" --start=2022-02-02 --end=2022-02-03 --ccf_path="~/ccf"
```
Result: No additional correlations compute because `2022-02-02-2022-02-03` is already considered 'Done'

Expected: Additional CCs are computed

carlosgjs:
A few different changes from learnings on AWS. A lot has to do with diagnosing and restarting/continuing long jobs:

- Better progress reporting via logging and progress bars (including memory)
- Use the `ConfigParameters.channels` field to pre-filter channels during load
- Up-fronts check for CCs that have already been computed, to avoid processing channels no longer needed (data load, preprocessing, FFTs)
- Remove CC Store `mark_done(timespan), is_done(timespan)` to allow for computing additional stations. This was a carry over from the old status text file. We now use `.contains(timespan, src, rec)` instead
- Update the CC Store `.contains()` method to be by station pair instead of channel pair.

Fixes #173


mdenolle (approved):
you must have dealt with a lot of bugs to write these functionalities...

PR #197, opened 2023-09-11
Update codecov in CI

niyiyu:
This PR updates the codecov wit current tests. Currently, the code coverage is about ~51%. We will add more tests to increase the coverage.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,7 +1,7 @@
 # config for codecov
 
-codecov:
-  token: eee2d1a8-f24b-401b-a211-19c10e987f53
+codecov: 
+  token: 5b43bc70-96b0-4bbc-abe3-e3c3584db624
 
 coverage:
   status:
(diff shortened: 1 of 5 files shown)


mdenolle (approved):
thanks for this improvement toward software best practices!

codecov.yml
```diff
@@ -1,7 +1,7 @@
 # config for codecov
 
 codecov:
-  token: eee2d1a8-f24b-401b-a211-19c10e987f53
+  token: df828ddb-54c2-4c3b-9959-0571877a1ea2
```
carlosgjs, line 4: @niyiyu is this needed here? I see that in the action you are fetching the token from a secret: `secrets.CODECOV_TOKEN`, which seems like the better way.

niyiyu: I had the same question when I was testing the token. I will test removing this and see if it pull the same token from the repo secret/variable, maybe in another PR.

PR #206, opened 2023-09-14
Add attenuation post processing

kuanfufeng:
This PR is adding scripts, notebook, and data for attenuation in the post processing.


niyiyu (approved):
Thanks!

src/noisepy/seis/application_modules/esyn_utils.py
```diff
@@ -0,0 +1,149 @@
...
+### -----
+# This scripts are aim to perform the 2-D radiative transfer equation
+# for scalar waves (Shang and Gao 1988; Sato 1993),
+# assuming of isotropic scattering and source radiation in infinite medium
+# to calculate synthesized energy densities Esyn.
+### -----
+
+
+### -----
+def read_pyasdf(sfile, ccomp):
```
carlosgjs, line 15: it would be better to add type annotations. Makes the code easier to use and understand
```suggestion
def read_pyasdf(sfile: str, ccomp: str) -> Tuple(float, float, np.ndarray, np.ndarray):
```

carlosgjs: To all functions

src/noisepy/seis/application_modules/esyn_utils.py
```diff
@@ -0,0 +1,149 @@
...
+    # useful parameters from each asdf file
+    with pyasdf.ASDFDataSet(sfile, mode="r") as ds:
+        alist = ds.auxiliary_data.list()
+        try:
+            dt = ds.auxiliary_data[alist[0]][ccomp].parameters["dt"]
```
carlosgjs, line 20: I wonder if the StackStore could be used for this instead of direct asdf file access? That way your function would work for any of the supported formats (asdf, zarr, numpy)

src/noisepy/seis/application_modules/esyn_utils.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+### -----
+# Function that Calculate Mean Square
+def msValue(arr, n):
+    square = 0
+    mean = 0.0
+
+    # Calculate square
+    for i in range(0, n):
```
carlosgjs, line 43: Can this be done with numpy? https://numpy.org/doc/stable/reference/generated/numpy.square.html. Something like:
```
square = np.sum(np.square(arr))
```

src/noisepy/seis/application_modules/esyn_utils.py
```diff
@@ -0,0 +1,149 @@
...
+    s0 = c**2 * tm**2 - r**2
+    if s0 > 0:
+        # second term
+        ind2 = mean_free ** (-1) * (math.sqrt(s0) - c * tm)
+        a2up = math.exp(ind2)
+        a2bot = 2 * math.pi * mean_free * math.sqrt(s0)
+        second = (a2up / a2bot) * step(tm - r / c)
+        Esyn = second
+
+        return Esyn
```
carlosgjs, line 85: what should this function return if `s0<=0`?

carlosgjs: same comment for next function

src/noisepy/seis/application_modules/esyn_utils.py
```diff
@@ -0,0 +1,149 @@
...
+        a2bot = 2 * math.pi * mean_free * math.sqrt(s0)
+        second = (a2up / a2bot) * step(tm - r / c)
+
+        Esyn = first + second
+
+        return Esyn
+
+
+### -----
+def convertTuple(tup):
```
carlosgjs, line 121:
```
"".join(tup)
```

src/noisepy/seis/application_modules/esyn_utils.py
```diff
@@ -0,0 +1,149 @@
...
+    if not np.ma.is_masked(arr):
+        med = np.median(arr)
+        data = np.median(np.abs(arr - med))
+    else:
+        med = np.ma.median(arr)
+        data = np.ma.median(np.ma.abs(arr - med))
+    return data
+
+
+### -----
```
carlosgjs, line 149:
```suggestion
from .noise_module import mad
```

carlosgjs (changes requested):

src/noisepy/seis/application_modules/esyn_utils.py
kuanfufeng: s0>0 makes the equation hold, so i did not consider getting negative value here. Do you have any suggestion for avoiding crushing?

src/noisepy/seis/application_modules/esyn_utils.py
kuanfufeng: Got it, thanks!

src/noisepy/seis/application_modules/esyn_utils.py
kuanfufeng: Updated!

src/noisepy/seis/application_modules/esyn_utils.py
kuanfufeng: It is a good point that using StackStore makes the function more practical. While the applications are all still currently relying on ASDF, maybe we can leave this to the next PR. What do you think?

carlosgjs: Yeah, there can a subsequent PR to update all of them

src/noisepy/seis/application_modules/esyn_utils.py
carlosgjs: you can raise an exception, e.g.
```
if s0 <= 0:
    raise ValueError(f"A message that makes sense to the caller, like c**2 * tm**2 - r**2  == {s0} <=0")
```

src/noisepy/seis/application_modules/esyn_utils.py
```diff
@@ -35,13 +36,10 @@ def read_pyasdf(sfile, ccomp):
 
 ### -----
 # Function that Calculate Mean Square
-def msValue(arr, n):
-    square = 0
+def msValue(arr: np.ndarray, n: int) -> Tuple[np.ndarray]:
```
carlosgjs, line 39: A Tuple is not needed when there's a single value being returned, eg:
```suggestion
def msValue(arr: np.ndarray, n: int) -> np.ndarray:
```
Same for a few of the functions below.

src/noisepy/seis/application_modules/esyn_utils.py
carlosgjs: Can you create an issue for this? It could be a good CodeUW task

src/noisepy/seis/application_modules/esyn_utils.py
```diff
@@ -0,0 +1,122 @@
...
+
+### -----
+def read_pyasdf(sfile: str, ccomp: str) -> Tuple[float, float, np.ndarray, np.ndarray]:
+    # useful parameters from each asdf file
+    with pyasdf.ASDFDataSet(sfile, mode="r") as ds:
+        alist = ds.auxiliary_data.list()
+        try:
+            dt = ds.auxiliary_data[alist[0]][ccomp].parameters["dt"]
+            dist = ds.auxiliary_data[alist[0]][ccomp].parameters["dist"]
+            print("working on %s (comp: %s) that is %5.2fkm apart. dt: %.3f " % (sfile, ccomp, dist, dt))
```
carlosgjs, line 23: Please use logging instead of printing. You can add this at the top:
```
import logging
logger = logging.getLogger(__name__)
```
```suggestion
            logger.info("working on %s (comp: %s) that is %5.2fkm apart. dt: %.3f " % (sfile, ccomp, dist, dt))
```

src/noisepy/seis/application_modules/esyn_utils.py
```diff
@@ -0,0 +1,122 @@
...
+            # read stacked data
+            sdata = ds.auxiliary_data[alist[0]][ccomp].data[:]
+
+            # time domain variables
+            npts = sdata.size
+            tvec = np.arange(-npts // 2 + 1, npts // 2 + 1) * dt
+            return dist, dt, tvec, sdata
+
+        except Exception:
+            print("continue! no %s component exist" % ccomp)
```
carlosgjs, line 33:
```suggestion
            logger.warning(f"continue! no {ccomp} component exist")
```
or maybe `logger.error`?
I also prefer the string interpolation style: https://www.programiz.com/python-programming/string-interpolation. That's more of a preference but is also what we've been using in the rest of the code.

carlosgjs (changes requested):
A few more suggestions, thx!

src/noisepy/seis/application_modules/esyn_utils.py
kuanfufeng: It may have other components to be used, so maybe `logger.warning` here should be fine(?).
Good advice on using string interpolation style. That makes it look straightforward. Thanks!

src/noisepy/seis/application_modules/esyn_utils.py
kuanfufeng: Issue created.

src/noisepy/seis/application_modules/esyn_utils.py
```diff
@@ -30,47 +33,61 @@ def read_pyasdf(sfile: str, ccomp: str) -> Tuple[float, float, np.ndarray, np.nd
...
+        c :        The Rayleigh wave velocity
+        r :        The distance between the source and receiver
+    ----------------------------------------------
+    step:      The Heaviside function
+    ----------------------------------------------
+    Return:
+        Esyn:   The synthetic energy density
+    ----------------------------------------------
+    """
+    r = r + 0.00000001  # to avoid the a2bot becomes zero
```
carlosgjs, line 90: nit: define a constant for the 0.0000001?

src/noisepy/seis/application_modules/esyn_utils.py
```diff
@@ -113,10 +138,247 @@ def ESYN_RadiaTrans_intersta(mean_free: float, tm: float, r: float, c: float) ->
         Esyn = first + second
 
         return Esyn
+    if s0 <= 0:
```
carlosgjs, line 141:
```suggestion
   else:
```

src/noisepy/seis/application_modules/esyn_utils.py
```diff
@@ -81,22 +98,30 @@ def ESYN_RadiaTrans_onesta(mean_free: float, tm: float, r: float, c: float) -> T
         Esyn = second
 
         return Esyn
+    if s0 <= 0:
```
carlosgjs, line 101:
```suggestion
    else:
```
You could also have this check at the beginning of the function as validation and then not have the check for `s0>0` since at that point you know that's the case. A structure like:

```
def f(x):
   if not valid(x):
       raise ValueError(f"Invalid x: {x}")
   
   return some calculation with x 
```

carlosgjs (approved):
Looking good, some minor suggestions

src/noisepy/seis/application_modules/esyn_utils.py
kuanfufeng: Sounds good. I will update it with this function.

PR #208, opened 2023-09-14
Refactor StackStore to write all stacks for a pair

carlosgjs:
- Changes to StackStore
   - Changed `append` and `read_stacks` to take and return `List[Stack]` instead of individual stack parameters
   - Remove `mark_done` `is_done`, we use the list of present station_pairs instead
 - Added `Stack` class (analogous to `CrossCorrelation`) and refactored common functionality into a `AnnotatedData` base class
 - Fixed check for `"all"`--> `StackMethod.ALL` to use the strongly typed enum
 - Added the 'numpy' format to the set of integration tests


niyiyu (approved):
Thanks!

mdenolle (approved):

PR #209, opened 2023-09-14
Fix caching in XMLChannelCatalog

carlosgjs:
Before (from Marine's run) ~10 mins:
```
2023-09-13 23:10:12,725 140060162046848 INFO utils.log_raw(): TIMING CC Main: 591.1919 secs. for get 756 channels
```
After (test run) ~2.8 mins:
```
2023-09-14 23:17:17,047 140072190446464 INFO utils.log_raw(): TIMING CC Main: 167.7949 secs. for get 753 channels
```


niyiyu (approved):
Interesting. I am surprised thatT he decorator limits the performance this much.

carlosgjs:
When no max is specified the default is 128, which is not unreasonable but is too small for our use case. An unlimited lru cache could also be very bad in other scenarios.

PR #216, opened 2023-09-20
CC: Memory and perf improvements

carlosgjs:
Two main changes:
- Move the processing of a timespan (e.g. day) into a function and use a new ThreadPoolExecutor every time for better memory cleanup
- Process the paths so the check for already done correlations can be done in O(1) time

Test processing 3 days show memory steady:
image


niyiyu (approved):
thank you !

PR #218, opened 2023-09-20
Process multiple timespans per node in array jobs

carlosgjs:
Query the parent job to figure out the array size. This way we can partition the days across all the array nodes.

This [example job](https://us-west-2.console.aws.amazon.com/batch/home?region=us-west-2#jobs/fargate/detail/c4bd3b5e-652d-4d08-bb5d-fba012a24869) ran 4 days on 2 nodes, where each node processed 2 days:

Node 1:
```
Process the chunk of 2022-02-05T00:00:00+0000 - 2022-02-06T00:00:00+0000
Process the chunk of 2022-02-03T00:00:00+0000 - 2022-02-04T00:00:00+0000
```

Node 2:
```
Process the chunk of 2022-02-06T00:00:00+0000 - 2022-02-07T00:00:00+0000
Process the chunk of 2022-02-04T00:00:00+0000 - 2022-02-05T00:00:00+0000
```


niyiyu (approved):
Looks cool. Thanks!

PR #219, opened 2023-09-21
Fix source dist testing

carlosgjs:
When testing the source distribution install, only install noisepy from source. Installing **all** from source it's unnecessary and more unstable. E.g. recently `hatchling==1.18.0` was failing to install from source.

The change to `gen_req.py` is opportunistic since it was missed in the last AWS PR and is only used in dev testing.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -82,7 +82,7 @@ jobs:
         echo "=== Testing source tar file ==="
         # Install tar gz and check import
         python -m pip uninstall --yes noisepy-seis
-        python -m pip install --extra-index-url https://test.pypi.org/simple --upgrade --pre --no-binary=:all: noisepy-seis==$latest_version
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yaml
```diff
@@ -82,7 +82,7 @@ jobs:
         echo "=== Testing source tar file ==="
         # Install tar gz and check import
         python -m pip uninstall --yes noisepy-seis
-        python -m pip install --extra-index-url https://test.pypi.org/simple --upgrade --pre --no-binary=:all: noisepy-seis==$latest_version
+        python -m pip install --extra-index-url https://test.pypi.org/simple --upgrade --pre --no-binary=noisepy-seis noisepy-seis==$latest_version
```
lsetiawan, line 85: Nice! That makes a whole lotta sense! 😄

lsetiawan (approved):

PR #224, opened 2023-09-27
Integration tests using SCEDC S3 data

carlosgjs:
Avoid depending on the FDSN download service, which is often flaky


niyiyu (approved):
Thank you! Looks great!

PR #230, opened 2023-09-28
Changes to store interfaces and implementations

carlosgjs:
See discussion: https://github.com/noisepy/NoisePy/discussions/229

Summary of changes:
- Redefined CrossCorrelationStore and StackStore to be more consistent
    - `read_correlations` and `read_stacks` are now both `read`
    - changes the order of parameters in the `contains()` methods
- Changed the implementations of the numpy and zarr stores to use the hierarchy described in the discussion. These now share the bulk of the implementation for both CC and Stack.
- Implemented an efficient caching of the cc/stack store directory

### DIFF
--- a/src/noisepy/seis/asdfstore.py
+++ b/src/noisepy/seis/asdfstore.py
@@ -2,7 +2,7 @@ import glob
 import logging
 import os
 from pathlib import Path
-from typing import Callable, Dict, Generic, List, Optional, Tuple, TypeVar
+from typing import Callable, Dict, Generic, List, Optional, Set, Tuple, TypeVar
 
 import numpy as np
 import obspy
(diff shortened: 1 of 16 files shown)


src/noisepy/seis/asdfstore.py
```diff
@@ -175,20 +183,39 @@ def read_correlations(self, timespan: DateTimeRange, src_sta: Station, rec_sta:
...
+        return f"{src_sta}_{rec_sta}"
+
 
 class ASDFStackStore(StackStore):
     def __init__(self, directory: str, mode: str = "a"):
         super().__init__()
         self.datasets = ASDFDirectory(directory, mode, _filename_from_stations, _parse_station_pair_h5file)
 
-    def append(self, src: Station, rec: Station, stacks: List[Stack]):
+    # TODO: Do we want to support storing stacks from different timespans in the same store?
```
mdenolle, line 206: probably

src/noisepy/seis/asdfstore.py
```diff
@@ -175,20 +183,39 @@ def read_correlations(self, timespan: DateTimeRange, src_sta: Station, rec_sta:
...
+    def append(self, timespan: DateTimeRange, src: Station, rec: Station, stacks: List[Stack]):
         for stack in stacks:
             self.datasets.add_aux_data((src, rec), stack.parameters, stack.name, stack.component, stack.data)
 
     def get_station_pairs(self) -> List[Tuple[Station, Station]]:
         return self.datasets.get_keys()
 
-    def read_stacks(self, src: Station, rec: Station) -> List[Stack]:
+    def get_timespans(self, src: Station, rec: Station) -> List[DateTimeRange]:
+        # TODO: Do we want to support storing stacks from different timespans in the same store?
```
mdenolle, line 215: yes eventually, we might want to store all timestamp from a given channel pair into the same store. this would be especially true for the "monitoring" application of this package.

mdenolle (approved):
What a major body of work!

src/noisepy/seis/asdfstore.py
carlosgjs: Sounds good, added https://github.com/noisepy/NoisePy/issues/232

PR #234, opened 2023-10-05
Added test for SingleNodeScheduler and fixed a bug in noise_module.py

### LINKED ISSUE #227: Write unit tests for the 3 schedulers
The 3 types in `scheduler.py` need unit tests.

- Familiarize yourself with [pytest](https://realpython.com/pytest-python-testing/)
- [test_remove_nan_rows](https://github.com/noisepy/NoisePy/blob/948d95f68d8120f23866c6442b502b4ae7e271af/tests/test_utils.py#L41) is a an example of a simple test
- Add a test file called `tests/test_schedulers.py`
- Add a test methods to test the 3 schedulers: `SingleNodeScheduler`, `AWSBatchArrayScheduler` and `MPIScheduler`. E.g. 
```
def test_single()
    pass

def test_aws()
    pass

def test_mpi()
    pass
```
- Run the tests on that file, and verify they pass:
```
pytest tests/test_schedulers.py
```
- Proceed to implement the 3 tests

## SingleNodeScheduler
Implement `test_single()` to call and verify the output of `initialize()`, `get_indices()` and `synchronize()`

## AWSBatchArrayScheduler
TBD

## MPIScheduler
TBD
```
import os
os.environ["OMPI_COMM_WORLD_SIZE"] = "3"
```
```
OMPI_COMM_WORLD_SIZE
OMPI_COMM_WORLD_RANK
```
https://docs.open-mpi.org/en/v5.0.x/tuning-apps/environment-var.html

IshikaKhandelwal:
resolves #227


mdenolle (approved):

tests/test_scheduler.py
```diff
@@ -0,0 +1,105 @@
+import os
+from unittest.mock import Mock, patch  # noqa: F401
+
+import pytest
+from mock import MagicMock
+
+from noisepy.seis.constants import AWS_BATCH_JOB_ID
```
carlosgjs, line 7:
```suggestion
from noisepy.seis.constants import AWS_BATCH_JOB_ID, AWS_BATCH_JOB_ARRAY_INDEX
```

tests/test_scheduler.py
```diff
@@ -0,0 +1,105 @@
...
+
+
+# get_array_size test
+@patch("boto3.client")
+def test_get_array_size(mock_boto3_client):
+    boto3_response = {"jobs": [{"arrayProperties": {"size": 10}}]}
+
+    # Test 1 - valid job id format
+    # Mock the AWS_BATCH_JOB_ID environment variable with a valid format
+    os.environ["AWS_BATCH_JOB_ID"] = "job_id:123"
```
carlosgjs, line 38: nit: You can assign `job_id:123` to a constant (e.g. `VALID_JOB_ID`) so that you don't have to hard code it 4 times

tests/test_scheduler.py
```diff
@@ -0,0 +1,105 @@
...
+    scheduler = AWSBatchArrayScheduler()
+    # Mock the _get_array_size method to return a known value
+    get_array_size_mock = MagicMock(return_value=5)
+
+    # Assign the MagicMock to the _get_array_size method of the class
+    AWSBatchArrayScheduler._get_array_size = get_array_size_mock
+
+    item = [1, 2, 3, 4, 5]
+    # indices = AWSBatchArrayScheduler.get_indices(item)
+    indices = scheduler.get_indices(item)
```
carlosgjs, line 77:
```suggestion
    items = [1, 2, 3, 4, 5]
    indices = scheduler.get_indices(items)
```

tests/test_scheduler.py
```diff
@@ -0,0 +1,105 @@
...
+
+    assert indices == [2]
+
+
+def test_get_indices_missing_index():
+    os.environ["AWS_BATCH_JOB_ARRAY_INDEX"] = ""
```
carlosgjs, line 83: No need to set this for the missing index case (see comment in `scheduler.py` at the top)
```suggestion
```

tests/test_scheduler.py
```diff
@@ -0,0 +1,105 @@
...
+    is_array = AWSBatchArrayScheduler.is_array_job()
+
+    assert is_array is True
+
+
+def test_is_array_job_without_index():
+    os.environ["AWS_BATCH_JOB_ARRAY_INDEX"] = ""
+    is_array = AWSBatchArrayScheduler.is_array_job()
+
+    assert is_array
```
carlosgjs, line 105: This should not return true if the index value is not valid. See suggestion above, I think we can cover both cases in one test.

tests/test_scheduler.py
```diff
@@ -0,0 +1,105 @@
...
+        scheduler.get_indices(items)
+    assert str(e.value) == "AWS_BATCH_JOB_ARRAY_INDEX environment variable not set"
+
+
+# is_array_job test
+def test_is_array_job_with_index():
+    os.environ["AWS_BATCH_JOB_ARRAY_INDEX"] = "2"
+    is_array = AWSBatchArrayScheduler.is_array_job()
+
+    assert is_array is True
```
carlosgjs, line 98:
```suggestion
    # no index
    assert not AWSBatchArrayScheduler.is_array_job() 
    # with index
    os.environ["AWS_BATCH_JOB_ARRAY_INDEX"] = "2"
    assert AWSBatchArrayScheduler.is_array_job()
```

tests/test_scheduler.py
```diff
@@ -0,0 +1,105 @@
...
+    os.environ["AWS_BATCH_JOB_ID"] = "job_id:123"
+    mock_boto3_client.return_value.describe_jobs.return_value = boto3_response
+    array_size = AWSBatchArrayScheduler._get_array_size()
+    assert array_size == 10
+
+    # Test 2 : Invalid job_id format
+    os.environ["AWS_BATCH_JOB_ID"] = "invalid_format"
+    with pytest.raises(ValueError) as e:
+        AWSBatchArrayScheduler._get_array_size()
+    assert str(e.value) == f"{AWS_BATCH_JOB_ID} is not in the expected format for an array job: 'invalid_format'"
```
carlosgjs, line 47: I suggest doing something like this:
```
assert "invalid_format" in str(e.value)
```
The reason is that having the full message has a minor drawbacks: It duplicates the error message between the code and the test and violates the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). This means that it the message is changed even a little bit, the test will break. 

By just asserting the presence of the `invalid_format` you remove the duplication and reduce the changes the test will need to be updated. But you still validate the main behavior, which is that there will be a ValueError about the 'invalid_format`.

You can do something similar with the other assertions below. E.g. `assert "123" in str(e.value)`

carlosgjs (changes requested):
Great progress! See my comments/suggestions. 

Once you address these you can move onto the MPIScheduler :)

tests/test_scheduler.py
IshikaKhandelwal: If I remove this line, pytest assertion fails with the message "Failed: DID NOT RAISE <class 'OSError'>"

.github/actions/setup/action.yaml
```diff
@@ -1,33 +1,25 @@
...
-  python-version:
-    required: true
-    description: Python version to install
-  mpi:
-    required: false
-    description: Whether to install with MPI or not
-    default: 'false'
-runs:
-  using: composite
-  steps:
```
carlosgjs, line 13: Looks like lines 1-13 got accidentally removed?

.github/actions/setup/action.yaml
```diff
@@ -1,33 +1,25 @@
...
+  run: pip install ".[dev,aws]"
+- name: Install project MPI
+  if: ${{ inputs.mpi == 'true' }}
+  shell: sh
+  run: pip install ".[dev,mpi,aws]"
+- name: Install Dependencies
+  run: pip install ".[dev,aws]"
+      if [[ ${{ inputs.mpi }} == 'true' ]]; then
+        pip install ".[mpi]"
+      fi
```
carlosgjs, line 25: I don't believe this is necessary. The `Install project MPI` and `Install project no MPI` should do this, right?

tests/test_scheduler.py
```diff
@@ -0,0 +1,182 @@
...
+import pytest
+
+from noisepy.seis.constants import AWS_BATCH_JOB_ARRAY_INDEX, AWS_BATCH_JOB_ID
+from noisepy.seis.scheduler import (
+    AWSBatchArrayScheduler,
+    MPIScheduler,
+    SingleNodeScheduler,
+)
+
+# from mpi4py import MPI
```
carlosgjs, line 15:
```suggestion
```

tests/test_scheduler.py
```diff
@@ -0,0 +1,182 @@
...
+    item = [1, 2, 3, 4, 5]
+    indices = scheduler.get_indices(item)
+
+    assert indices == [2]
+
+    os.environ.pop(AWS_BATCH_JOB_ARRAY_INDEX, None)
+
+
+def test_get_indices_missing_index():
+    # os.environ["AWS_BATCH_JOB_ARRAY_INDEX"] = ""
```
carlosgjs, line 107:
```suggestion
```

carlosgjs (approved):
Great! Please see two small inline suggestions (remove commented code) before merging.

PR #235, opened 2023-10-09
Stacking fixes/improvements

carlosgjs:
- Standardize and validate dates to be UTC
- Rename `stack()` function to `stack_cross_correlations()` to avoid name conflict with the `stack.py` module. @mdenolle another option is to rename the module. E.g to `stacking.py` and keep the function as `stack()`. WDYT?
- Parallelize data loading in `plot_all_moveout `. 
    - RFC: Support sub-sampling a portion of the station pairs for faster plotting (see comment inline)
- Add simple notebook for plotting existing stacked data
- Fix loading of config from S3 bucket
- Change default format to NUMPY

### DIFF
--- a/src/noisepy/seis/plotting_modules.py
+++ b/src/noisepy/seis/plotting_modules.py
@@ -1,5 +1,7 @@
 import logging
 import os
+import random
+from concurrent.futures import ThreadPoolExecutor
 
 import matplotlib.pyplot as plt
 import numpy as np
@@ -11,6 +13,7 @@ from obspy.signal.filter import bandpass
 from scipy.fftpack import next_fast_len
 
 from noisepy.seis.stores import CrossCorrelationDataStore, StackStore
+from noisepy.seis.utils import TimeLogger, get_results
 
 logging.getLogger("matplotlib.font_manager").disabled = True
 logger = logging.getLogger(__name__)
@@ -634,6 +637,7 @@ def plot_all_moveout(
     freqmax,
     ccomp,
     dist_inc,
+    sample_pairs=1.0,
     disp_lag=None,
     savefig=False,
     sdir=None,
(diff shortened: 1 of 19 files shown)


src/noisepy/seis/plotting_modules.py
```diff
@@ -690,27 +694,35 @@ def plot_all_moveout(
     indx1 = int((maxlag - disp_lag) / dt)
     indx2 = indx1 + 2 * int(disp_lag / dt) + 1
 
+    if sample_pairs < 1.0:
```
carlosgjs, line 697: @mdenolle @niyiyu @kuanfufeng - I'd like your input here. When testing this plot on a large number of station pairs (e.g. 40k) it can take a long time. So I added a parameter to sub-sample and only plot a portion of the available stations. Here are the times when running a notebook on AWS:
```
5% sampling -> 1.85 mins
25% sampling -> 9.4 mins
100%  (no) sampling -> 37 mins
```
This helps when debugging but I don't if it makes any sense to do so when analyzing the data. I'm happy to take it out or maybe there's a better way to sub-sample instead of random? The problem is that reading the metadata is just as expensive as reading the data.

src/noisepy/seis/plotting_modules.py
```diff
@@ -690,27 +694,35 @@ def plot_all_moveout(
...
             logger.warning(f"No data available for {src}_{rec}/{stack_name}/{ccomp}")
-            continue
+            return
         params, all_data = stacks[0].parameters, stacks[0].data
         dist[ii] = params["dist"]
         ngood[ii] = params["ngood"]
         crap = bandpass(all_data, freqmin, freqmax, int(1 / dt), corners=4, zerophase=True)
         data[ii] = crap[indx1:indx2]
 
+    tlog = TimeLogger(level=logging.INFO)
```
carlosgjs, line 720: This helps bring the time down to 37 mins from almost 7hrs!

niyiyu (approved):
Thanks!

PR #244, opened 2023-10-16
refactor: Decouple data loading from plot_all_moveout

carlosgjs:
- Moved the parallel data loading code from `plot_all_moveout` to a helper on the store object
- Updated parameters and notebooks
- Unrelated: fixed lint about using `isinstance` instead of `type(...)==`

This makes it easier for someone to experiment what to plot. As a simple example

## Loading only 10% of the stacks from 2022 (2 mins)
image

## For reference, here's with 100% of the data (47 mins):
image


niyiyu (approved):
Thanks!

PR #289, opened 2024-02-12
Added a dependency on the new noisepy-seis-io and fixed the import statements

IshikaKhandelwal:
Added a dependency on the `noisepy-seis-io` and fixed the import statements. Moved `channel_filter_store.py `and `test_channel_filter_store.py` to `noisepy-seis-io`. Created a mock `MockCatalog` class to be utilized in `test_crosscorrelation.py`


.vscode/settings.json
```diff
@@ -3,7 +3,7 @@
     "editor.defaultFormatter": "ms-python.black-formatter",
     "editor.formatOnSave": true,
     "editor.codeActionsOnSave": {
-      "source.organizeImports": true
+      "source.organizeImports": "explicit"
```
carlosgjs, line 6: what does this do?

pyproject.toml
```diff
@@ -50,6 +50,7 @@ dependencies = [
     "PyYAML==6.0",
     "pydantic-yaml==1.0",
     "psutil>=5.9.5,<6.0.0",
+    "noisepy-seis-io",
```
carlosgjs, line 53: I recommend adding a minimum version (e.g. `>=0.1.0`) once the noisepy-seis-io package is published

carlosgjs (dismissed):
The tutorial notebooks need to be updated too, right?

IshikaKhandelwal:
> The tutorial notebooks need to be updated too, right?

Yes, I plan to do that in the next PR. For now it would be just all the changes in the import statements

.vscode/settings.json
IshikaKhandelwal: The configuration for organizing imports in our code editor changed automatically from true to "explicit". This likely occurred due to an update or default setting adjustment within the development environment.

.github/actions/setup/action.yaml
```diff
@@ -26,8 +26,14 @@ runs:
     - name: Install project no MPI
       if: ${{ inputs.mpi == 'false' }}
       shell: sh
-      run: pip install ".[dev,aws]"
+      run: |
+        pip install noisepy-seis-io --no-deps --target tmp
```
carlosgjs, line 30: I worry this method is likely hiding an issue a user might face. It should be possible for the `noisepy-seis-io` package to be pulled in like any other dependency. E.g. we don't have to install any other packages like `numpy` or `pandas` like this,.

.github/workflows/test.yaml
```diff
@@ -30,6 +30,11 @@ jobs:
     steps:
     - name: Checkout Repo
       uses: actions/checkout@v3.3.0
+
+    - name: Install noisepy-seis-io
```
carlosgjs, line 34: We shouldn't need this. If we do it's indicative of a problem with having it as a dependency.

carlosgjs (changes requested):

mdenolle (approved):

PR #297, opened 2024-02-21
feat: Add substack_windows config parameter

carlosgjs:
@mdenolle , this is how I would suggest addressing issue noisepy/noisepy-io#36 .

However, it would be better to do this once Ishika finishes her refactoring because `datatypes.py` has moved to the `noisepy-io` repo.

@IshikaKhandelwal - fyi

### DIFF
--- a/src/noisepy/seis/datatypes.py
+++ b/src/noisepy/seis/datatypes.py
@@ -193,9 +193,8 @@ class ConfigParameters(BaseModel):
     substack: bool = Field(
         default=False, description="True:  smaller stacks within the time chunk. False: it will stack over inc_hours"
     )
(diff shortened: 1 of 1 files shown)


src/noisepy/seis/datatypes.py
```diff
@@ -274,8 +277,6 @@ def validate_date(d: datetime, name: str):
 
         validate_date(m.start_date, "start_date")
         validate_date(m.end_date, "end_date")
```
carlosgjs, line 279: no longer needed since now substack_len is computed

niyiyu:
Addressing in IO package. https://github.com/noisepy/noisepy-io/pull/37

PR #298, opened 2024-02-21
chore: moving unused functions to improve code coverage

niyiyu:
There are a lot of functions that are not used in the new NoisePy, from the Jiang and Denolle 2019. I am putting them into an isolated directory at src/noisepy/old so that codecov test is to preformed on these uncovered codes.

- This should improve the codecov of noise_module as mentioned in #250.
- I am also adding more tests for stack and cross-correlation module.


mdenolle (commented):
Could you rename "old" to "functions_2019" . Could you explain why iflag is left commented in your PR?

niyiyu:
> Could you rename "old" to "functions_2019" .

Yes

> Could you explain why iflag is left commented in your PR?

There is a problem here. `iflag` checks if we got CCF in every component pair. And it checks all 9 components: EE, EN, EZ, NE, NN, NZ, ZE, ZN, ZZ. However, we only save the independent 6 component so that the rest 3 will always be empty, making iflag ending with 0. In the following codes where `iflag` is used, the rotation is never performed.

mdenolle (approved):

PR #320, opened 2024-09-27
Update attenuation modules

kuanfufeng:
This PR is for updating attenuation modules with the multi-window option.
- The original `get_SSR` is renamed as `get_codawindow_SSR`.
- The new module, `get_codawindow_SSR`, can handle both single and multiple windows cases. 
- Another new module, `window_determine`, slices the input coda window into ten small windows and returns the first 30% and the 30% after a half point to the call. 
- By now, the multi-window number is fixed at 6, according to the article setting (Hirose et al., 2019, 2022).

There is one more modification, regardless of the monitoring module. For correction, the parameter `smooth_N` is replaced by `smoothspect_N` in noise_module.


src/noisepy/monitoring/attenuation_utils.py
```diff
@@ -136,20 +137,202 @@ def check_s0(x: float) -> None:
...
+        mean_free_path: The mean free path
+        intrinsic_b: The intrinsic absorption parameter b
+        npts: The total sample number
+    ----------------------------------------------
+    Return:
+        window_SSR: The sum of squared residuals in the specific window
+    """
+    Esyn_temp = np.zeros((npts // 2))
+    Eobs_temp = np.zeros((npts // 2))
+    PLOT_CHECK = False
```
niyiyu, line 271: Considering moving this as the function arguments.

src/noisepy/monitoring/attenuation_utils.py
```diff
@@ -136,20 +137,202 @@ def check_s0(x: float) -> None:
...
+def windowing_SSR(
+    coda_window: np.array,
+    mean_free_path: float,
+    intrinsic_b: float,
+    fmsv: np.ndarray,
+    dist: np.ndarray,
+    npts: int,
+    dt: float,
+    cvel: float,
+    ncoda: int,
```
niyiyu, line 253:
```suggestion
    ncoda: int,
    PLOT_CHECK: bool = False,
```

src/noisepy/monitoring/monitoring_utils.py
```diff
@@ -50,6 +50,7 @@ class ConfigParameters_monitoring(BaseModel):
     ratio: float = Field(default=3.0, description="ratio for determining noise level by Mean absolute deviation (MAD)")
 
     # --- paramters for measuring attenuation
+    single_station: bool = Field(default=True, description="make measurement one one lag or two")
```
niyiyu, line 53:
```suggestion
    single_station: bool = Field(default=True, description="make measurement on one lag or two")
```

niyiyu (commented):
Thanks. There are just two very minor comments.

src/noisepy/monitoring/monitoring_utils.py
kuanfufeng: Thank you for noticing this typo. I would need to update the description of this.

src/noisepy/monitoring/attenuation_utils.py
kuanfufeng: Thanks for the suggestion. I will add the `PLOT_CHECK` as a function argument. `Esyn_temp` and `Eobs_temp` will keep recalculating via get_energy_density based on the parameters. Maybe we can leave them as they are.

Issue #352, opened 2026-05-21
Irregular time steps in CCF


koepflma:
Hello noisepy-team
I'm currently using noisepy to calculate correlation functions. I recognized, that the time array in file saved in CCF is irregular. In my case, I use step = 300 seconds, so I would expect that the difference between two consecutive values in attrs["time"] is 300, instead it varies between 256 and 384. I assume this comes from initiating dataS_t as float32 array, but float64 would be needed.

I hope this code snippet and the provided file help to understand the issue.

Thank you for your help!
Manuela

```
with h5py.File("./CCF/2025_08_07_00_00_00T2025_08_08_00_00_00.h5", "r") as f:
    print(f["AuxiliaryData/Z5.002_Z5.002/DPZ_DPZ"].shape)
    attrs = dict(f["AuxiliaryData/Z5.002_Z5.002/DPZ_DPZ"].attrs)

np.diff(attrs["time"])
```

[2025_08_07_00_00_00T2025_08_08_00_00_00.h5.zip](https://github.com/user-attachments/files/28110442/2025_08_07_00_00_00T2025_08_08_00_00_00.h5.zip)

PR #356, opened 2026-08-03
Performance fixes from the 2026-08 audit (float32 windows, batched irfft, rfft whitening) — timestamps stay float64

mdenolle:
## Summary

Performance fixes from the 2026-08 profiling audit of the S1 cross-correlation pipeline, in one branch as requested. The headline change reverts the waveform window matrix to float32 while **keeping timestamps float64** — please read the context below.

### Context: issue #352 / PR #353 (please review, @niyiyu; cc @koepflma (Manuela Koepfli))

Issue #352 (reported by Manuela Koepfli) showed irregular CCF `time` attributes: 256/384 s increments instead of the configured 300 s step. The root cause was that the timestamp array `dataS_t` was float32 — a float32 unix epoch at current dates only resolves ~256 s. PR #353 (Yiyu Ni) correctly fixed this by moving window cutting to float64.

**That fix is preserved and is now locked in by a regression test**: `dataS_t` stays float64, and `tests/test_noise_module.py::test_cut_trace_make_stat_timestamps_are_regular` builds timestamps at a 2025 epoch with step=300 and asserts `np.diff(dataS_t)` is exactly 300 — so the #352 fix can never silently regress. The same commit also found and fixed one remaining float32 timestamp array in `correlate()`'s `substack_len != cc_len` branch (`t_corr` was still `np.float32`).

However, PR #353 also flipped the waveform window matrix `dataS` to float64, which was not needed for #352 and doubles the memory and FFT cost of everything downstream. This PR takes `dataS` back to float32. A numerical check at campaign geometry (fs=40 Hz, cc_len=1800 s, 188 windows, RMA whitening, ±32 s lags) shows the float32 waveform path recovers CCFs with correlation r=0.9999996 against float64, residuals at the 1e-4-of-peak level, and identical dv/v from a stretching grid search.

### Changes

1. **`dataS` back to float32** (`cut_trace_make_stat`); `dataS_t` stays float64, with the regression test above.
2. **Batched hermitian irfft in `correlate()`**: the per-window loop of nwin individual `scipy.fftpack.ifft(Nfft)` complex calls is replaced with one `scipy.fft.irfft(axis=1, workers=-1)` over the half spectra (the manually mirrored spectrum was hermitian with a zero Nyquist bin, which is exactly irfft's contract). The frequency-domain mean removal now accumulates in float64. This also removes the GIL bottleneck that limited `correlate.py`'s ThreadPoolExecutor scaling (audit: 0.64 s → 0.04 s per pair-day; ~1.7x → near-linear on 4 threads).
3. **Dead allocations removed**: the unused `arr_out` in `whiten()` (433 MB transient in the 2D branch at campaign geometry — allocated, filled, and discarded) and the dead `np.zeros(nwin * Nfft2)` in `correlate()` (54 MB, immediately overwritten).
4. **`whiten_2D` uses `rfft`** instead of a full complex `fftn`: the legacy code zeroed everything outside the whitened positive-frequency band anyway, so only that band needs to be computed and filled. Output shape and values are unchanged. **`moving_ave_2D`** now uses `scipy.ndimage.uniform_filter1d`, verified numerically identical (max diff ~1e-14) to the padded `convolve2d` for both odd and even window lengths.
5. **`detrend` vectorized**: the per-row loop becomes two matrix products (same QR-based operator, in-place update so dtype is preserved; bitwise-identical in the benchmark). The redundant `mad() != 0` gate in `cut_trace_make_stat` is dropped — the `std` check covers it and `mad()` is a full O(n log n) median over the day-long trace.
6. **Autocorrelation detection**: `correlate()` now takes an `is_autocorr` flag derived from `iiS == iiR` at the call site instead of comparing the two full amplitude spectra (~25 ms per pair). The old comparison is kept as a fallback when the flag isn't passed (external callers), but note it was only ever reliable for the XCORR method — for DECONV/COHERENCY the smoothed source spectrum never compares equal, so autocorrelations were misclassified as cross-correlations there; the explicit flag is semantically correct for all methods.

### A/B numbers (full-pipeline, seeded synthetic day, RMA/RMA, XCORR)

From `tests/test_ccf_regression.py`, which runs the complete legacy float64 path (per-window `fftpack.ifft`, `fftn` whitening, `convolve2d` smoothing — verbatim copies of the pre-change code) against the new path:

| Metric | Value |
|---|---|
| per-window CCF correlation (22 windows) | min 0.999999881 |
| stacked CCF correlation | 1.000000000 (9 decimals) |
| max stacked residual / peak | 2.5e-8 |
| stretching grid search (±5%, legacy vs new stack) | dv/v = -0.002% (grid resolution 0.004%) |
| dv/v of a 0.5%-stretched target vs either stack | identical to machine precision (diff = 0.0) |
| timestamps | identical between paths, exactly regular (Δt = step) |

Micro-benchmarks at campaign geometry (188×72000, Apple Silicon):
- `correlate()` inverse FFTs: 0.39 s → 0.11 s
- `whiten_2D`: 0.74 s → 0.10 s
- `moving_ave_2D` (188×5000, N=10): 31 ms → 3 ms
- `detrend`: 161 ms → 94 ms (identical output)

### Tests

- New: `tests/test_ccf_regression.py` (A/B full-pipeline guard, substack and no-substack branches) and `tests/test_noise_module.py::test_cut_trace_make_stat_timestamps_are_regular` (#352 guard).
- All existing tests pass locally (`pytest tests/. integration_tests/.`).

@niyiyu and @koepflma — since the precision change touches the ground your #353 fix covered, requesting your review in particular: timestamps remain float64 everywhere (now including the substack branch), and the float32 change is confined to the waveform windows with the A/B numbers above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


niyiyu (approved):
Agreed that `dataS` is not necessary to be `np.float64`. Merging now.

